PR #9, opened 2017-11-21
Added HMR, created separate run target to make it faster

AkosLukacs:
Previously "Run" command called "Build", and created webpack production bundle.
Some dependencies updated to match the BookStore.


theimowski:
thanks!

PR #19, opened 2017-12-11
Rename `path` in Server/Program.fs to avoid collision with Suave

### LINKED ISSUE #18: Suave's `path` gets overriden by example code
There's this novice-unfriendly annoyance. The `path` definition in `Server/Program.fs`

    let path = Path.Combine("..","Client") |> Path.GetFullPath

covers Suave's own `path` function, so trying to do something like `choose [ (path "/") >=> OK "Hi" ]` won't work because by then `path` is a string rather than function *(causing excessive head scratching and wondering why Suave example code just won't work)*.

pkese:
trying to fix #18

### DIFF
--- a/Content/src/Server/Program.fs
+++ b/Content/src/Server/Program.fs
@@ -7,12 +7,12 @@ open Suave.Operators
 
 open Shared
 
-let path = Path.Combine("..","Client") |> Path.GetFullPath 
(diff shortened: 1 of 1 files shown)


theimowski:
Thanks, this is reasonable

PR #22, opened 2017-12-14
Remoting refactorings

Zaid-Ajaj:
This PR adds  (somewhat opinionated) refactorings to how Fable.Remoting should be used. Consider the code snippet:
```fs
    let api = Fable.Remoting.Client.Proxy.createWithBuilder<Init> routeBuilder
    Cmd.ofAsync 
      api.getCounter
      () 
      (Ok >> Init)
      (Error >> Init)
```
It suggests that the user needs to create a new proxy object whenever he or she wants to talk to server. But the idea is that you create the proxy once and forget about it like this, in it's own module:
```fs
module Server = 
  open Shard
  open Fable.Remoting.Client
  /// A proxy you can use to talk to server directly
  let api : ICounterProtocol = 
    Proxy.createWithBuilder<ICounterProtocol> routeBuilder
```
and to be used like this:
```fs
Cmd.ofAsync 
  Server.api.getInitCounter
  () 
  (Ok >> Init)
  (Error >> Init)
```
Another nice thing to have is sharing the code for `routeBuilder` and moving it to the `Shared` module to make sure that client and server are using the exact same route builder.

At last, I didn't think the name `Init` was appropriate for the protocol specification because it might suggest for new users that you need a new record type for each function. I renamed it to `ICounterProtocol` so that the users know they can put all the function definitions there (related to `Counter` operations). Personally I usually use `IClientServerProtocol` but that is too general in this case :smile:


theimowski:
This makes perfect sense, thanks!

theimowski:
Released in 0.2.2

Zaid-Ajaj:
@theimowski Thanks :smile:

PR #26, opened 2017-12-24
Run paket restore/install based on existence of paket.lock file

Zaid-Ajaj:


theimowski:
I don't think it's good idea to trigger `paket install` on every build. Ideally we'd trigger `paket install` after the template is created, but not yet sure how to achieve that

Zaid-Ajaj:
Solved, see last commit :smile:

Zaid-Ajaj:
@theimowski I just checked, ionide with vscode actually restores the packages right away when you open the project repo

theimowski:
Awesome - makes things simple. Added same to `build.cmd` and released in 0.4.2, thanks

PR #37, opened 2018-02-23
Reinstate postActions to make scripts executable on non-windows OSs

mtnrbq:
Tested both positive and negative with change detailed above by @rmunn


theimowski:
works like a charm, thanks! Released in 0.8.4

Issue #39, opened 2018-02-23
Build script may not work out of the box because of the Github crypto policy changes


psfinaki:
Git has just [changed](https://githubengineering.com/crypto-removal-notice/) something in their cryptographic standards policy. Because of that Paket [may not](https://github.com/fsprojects/Paket/issues/3065) automatically download. Consequently, the whole build.sh/build.cmd will fail right in the beginning (this was my case).

The simple [workaround](https://github.com/fsprojects/Paket/issues/3065#issuecomment-367829309) fixed the whole thing. Nevertheless I believe some of the SAFE investigators can get repelled not seeing stuff automagically running.

I propose some note about this in the docs until the issue there is resolved in some proper way.

isaacabraham:
It's something that's affecting more than just SAFE projects - basically any project that uses Paket is potentially affected by this. I'm not sure what the best solution for it is currently, but once the bootstrapper in the individual repositories have been replaced with the "new" bootstrapper, things will once again work.

psfinaki:
I have created a pull request to add a temp note to the README, just in case somebody else in the world gets suddenly interested in SAFE stack during this weekend. 

Not that important for me now, I got my SAFE up & running :)

landy:
I have created [pull request](https://github.com/SAFE-Stack/SAFE-template/pull/38) with updated paket version which should fix this problem

psfinaki:
@landy You have definitely done a better job :)

theimowski:
Should be now working with 0.8.3

PR #40, opened 2018-02-23
Updated README, note about Paket not downloading (#39)

psfinaki:
A temp note until they fix the issue out there


psfinaki:
@forki Great, I did not expect the fix to come that fast. I will not regret if you abandon this :D

forki:
@theimowski can you please release the updated template?

theimowski:
thanks, released

Issue #42, opened 2018-02-25
[NRE] It happens with Giraffe and NPM template


t-smirnov:
NRE happens here:
```f#
 task {
        let! counter = getInitCounter()
        >>> return! Successful.OK counter next ctx
    }
```
here's some stacktrace:
```
Exception has occurred: CLR/System.NullReferenceException
Exception thrown: 'System.NullReferenceException' in FSharp.Core.dll: 'Object reference not set to an instance of an object.'
   at Microsoft.FSharp.Core.LanguagePrimitives.IntrinsicFunctions.UnboxGeneric[T](Object source)
   at Giraffe.Negotiation.HttpContext.NegotiateAsync(HttpContext this, Object responseObj)

```

![nre-fsharp-safe-template](https://user-images.githubusercontent.com/17054766/36641700-4aa25f38-1a45-11e8-9a20-858b6576c4d8.PNG)

forki:
/cc @dustinmoris

Am 25.02.2018 14:03 schrieb "Tim" <notifications@github.com>:

t-smirnov:
additionally it occurs with latest netcore, win10, without VS installation

theimowski:
I can reproduce, looks to me like newest Giraffe 1.1.0 has issues with unboxing integer type for the response in: `return! Successful.OK 42 next ctx` - waiting for @dustinmoris to comment

dustinmoris:
Hi,

I've tested the following with the latest Giraffe 1.1.0 and it worked perfectly fine for me:

```fsharp
route "/int1" >=> Successful.OK 42
route "/int2" >=> (fun next ctx -> task { return! Successful.OK 42 next ctx })
```

My guess is that you haven't registered the default Giraffe dependencies which make sure that you have a valid JSON and XML serializer as dependencies:

```fsharp
let configureServices (services : IServiceCollection) =
    // Add Giraffe dependencies
    services.AddGiraffe() |> ignore
```

With Giraffe 1.0.0 I've moved away from `customJson` and other custom this and that functions in favour of registering serializers through ASP.NET Core's service collection, which makes it possible to configure one JSON serializer or one XML serializer for an entire Giraffe application, so that it works across the board, including `negotiate` and the status response helpers like `Successful.OK`.

It's been documented here: https://github.com/giraffe-fsharp/Giraffe/blob/master/DOCUMENTATION.md#plugging-giraffe-into-aspnet-core

[...]

PR #43, opened 2018-02-25
fixed template with Giraffe Server

t-smirnov:
Fixed [Issue42](https://github.com/SAFE-Stack/SAFE-template/issues/42) when Giraffe Server fails with NRE,  cause hasnt `ConfigureServices` section in Server.fs


theimowski:
thanks! Fix released in 0.8.5

PR #50, opened 2018-03-26
Add Fable Json serialization to Saturn and Giraffe.

### LINKED ISSUE #49: Add support for Fable JSON serialization
I think that if using either Giraffe or Saturn, the templates should include the Fable.Json package and have them auto-configured as part of the template.

Happy to help with this if you agree that this is worth doing!

isaacabraham:
@theimowski can you have a quick look at what I've done in this PR and see if it looks about right? I don't even know how to build the template locally so this is just being done "blind".

When complete, this will fix #49

### DIFF
--- a/Content/src/Server/paket.references
+++ b/Content/src/Server/paket.references
@@ -1,5 +1,6 @@
 group Server
   FSharp.Core
+
 //#if     (Server == "suave")
   Suave
 //#elseif (Server == "giraffe")
@@ -9,12 +10,15 @@ group Server
 //#elseif (Server == "saturn")
   Saturn
 //#endif
+
 //#if     (!Remoting && Server != "suave")
   Fable.JsonConverter
 //#endif  
+
 //#if     (Remoting && Server == "suave")
   Fable.Remoting.Suave
 //#elseif (Remoting)
   Fable.Remoting.Giraffe
 //#endif
+
   Microsoft.DotNet.Watcher.Tools
--- a/Content/src/Server/paket.references
+++ b/Content/src/Server/paket.references
@@ -14,9 +14,7 @@ group Server
 //#endif  
 //#if     (Remoting && Server == "suave")
   Fable.Remoting.Suave
-//#elseif (Remoting && Server == "giraffe")
-  Fable.Remoting.Giraffe
-//#elseif (Remoting && Server == "saturn")
+//#elseif (Remoting)
   Fable.Remoting.Giraffe
 //#endif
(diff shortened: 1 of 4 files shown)


theimowski:
Looks ok at first glance - you should be able to build using the build scripts in root directory, then add local template with `dotnet new -i nupkg/SAFE.Template.X.X.X.nupkg` (see [Contribution guide](https://github.com/SAFE-Stack/SAFE-template/blob/master/CONTRIBUTING.md#testing-nuget-package))

isaacabraham:
OK, I think this is ready for review / test. I tested it by doing the following:

1. Created a new project using the template using `-S saturn`.
2. Changed the type of `Counter` to `{ Value : int }`.
3. Fixed any errors in update:

```fsharp
let update msg (model : Model) =
  let model' =
    match model,  msg with
    | Some x, Increment -> Some { Value = x.Value + 1 }
    | Some x, Decrement -> Some { Value = x.Value - 1 }
    | None, Init (Ok x) -> Some x
    | _ -> None
  model', Cmd.none
```

4. At this point, the app should work. However, if you hit `api/init` directly, the call will fail because it will try to serialize to XML (I assume when Fable makes the client-side call, it requests a JSON response explicitly). So, you can do one of two things:
   a. You can attach the `[<CLIMutable>]` attribute to `Counter` so that it serializes correctly.
   b. You can make the following change in `apiRouter`:

```fsharp
      return! Successful.OK counter next ctx
```
with
```fsharp
      return! json counter next ctx
```

Content/src/Server/paket.references
```diff
@@ -9,11 +10,15 @@ group Server
...
+//#if     (!Remoting && Server != "suave")
+  Fable.JsonConverter
+//#endif  
+
 //#if     (Remoting && Server == "suave")
   Fable.Remoting.Suave
-//#elseif (Remoting && Server == "giraffe")
-  Fable.Remoting.Giraffe
-//#elseif (Remoting && Server == "saturn")
+//#elseif (Remoting)
```
isaacabraham: This removes a warning during template generation that added the same dependency twice.

theimowski:
Is this only applicable for Saturn option?

isaacabraham:
Mmm, good point. I suppose it did also be there for Giraffe as well. No clue how this works with Suave though.

isaacabraham:
@theimowski now added to Giraffe as well.

isaacabraham:
I *think* that this is good to go.

theimowski:
cool, thanks! released in 0.11

PR #52, opened 2018-04-04
Fixes path to index.html in Saturn template

jeremyabbott:
"/index.html" resolves to the machine root. "index.html" combined with clientPath works as expected.


theimowski:
Thanks!

PR #53, opened 2018-04-07
Update Remoting, use latest API

Zaid-Ajaj:
Update server adapters to 2.6 / client to 2.4, these are most stable and most heavily tested versions. 

Use latest `remoting` API to create the handlers in accordance with the [docs](https://zaid-ajaj.github.io/Fable.Remoting/), (still backwards compatible)   

I only tested the template with Suave server locally.

Suggestion: 
 - Use default remoting implementation without `Route.builder`?
 - rename `Route.builder` to `Route.prefixWithApi`?

### DIFF
--- a/Content/paket.dependencies
+++ b/Content/paket.dependencies
@@ -16,11 +16,11 @@ group Server
   nuget Fable.JsonConverter
 //#endif  
 //#if     (Remoting && Server == "suave")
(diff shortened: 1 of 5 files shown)


Zaid-Ajaj:
Ah ofcourse, I forgot :sweat_smile: otherwise the config would become `/ICounterProtocol/*` and you have to maintain it with every other protocol/change

PR #55, opened 2018-04-13
Link to safe docs

forki:


theimowski:
Yep that's a good idea to have that in one place. One thing I'm concerned about is updating the usage - every time we change something in the template, we should also remember to update the docs kept in separate template. Do you think this should be somehow automated?

theimowski:
For now I'll just make PRs to the docs whenever params change

PR #58, opened 2018-04-16
Add saturn cli if saturn selected.

jeremyabbott:


jeremyabbott:
Actually closing this because there are a few other things that would need to be added for the CLI to make sense here.

PR #59, opened 2018-04-18
Add charset attribute to index.html

marcpiechura:
I had to set the charset to utf8 in order to display `str “äöü“` correctly, otherwise it fallbacks to ascii?! which looks quite ugly ;-)

### DIFF
--- a/Content/src/Client/index.html
+++ b/Content/src/Client/index.html
@@ -2,15 +2,13 @@
 <html>
 <head>
     <title>SAFE Template</title>
+    <meta charset="utf-8">
+    
     <!--#if (Fulma != "none")
-    <meta charset="utf-8" name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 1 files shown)


Content/src/Client/index.html
```diff
@@ -3,12 +3,16 @@
 <head>
     <title>SAFE Template</title>
     <!--#if (Fulma != "none")
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta charset="utf-8" name="viewport" content="width=device-width, initial-scale=1">
```
forki: I think I did that recently in my code as well and html validation said it should be in a separate meta tag.

marcpiechura:
@forki addressed your remarks, also closed the elmish div correctly, according to https://validator.w3.org \<div /> isn't valid too.

theimowski:
Thanks! released in 0.13.1

PR #64, opened 2018-04-26
Put all static content inside the public dir

theprash:
This simplifies things a little bit in preparation for adding Azure deployment.


isaacabraham:
Can someone test this out to ensure that the docker deploy still works? This is a pre-req for #65 which will allow a pure Azure PAAS deployment.

isaacabraham:
@theprash I've tested it out and you can remove the `browserRouter` scope completely. Kestrel defaults to accepting index.html - see [here](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/static-files?view=aspnetcore-2.1&tabs=aspnetcore2x).

theimowski:
I've tested Saturn and Suave both with `build run` and docker images - looks good! Thanks

PR #67, opened 2018-04-27
Remove Saturn browser router

isaacabraham:
No longer needed.

### DIFF
--- a/Content/src/Server/ServerSaturn.fs
+++ b/Content/src/Server/ServerSaturn.fs
@@ -19,10 +19,6 @@ let port = 8085us
 
 let getInitCounter () : Task<Counter> = task { return 42 }
 
-let browserRouter = scope {
-  get "/" (htmlFile (Path.Combine(publicPath, "index.html")))
-}
-
 #if (Remoting)
 let server =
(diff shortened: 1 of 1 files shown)


Content/src/Server/ServerSaturn.fs
```diff
@@ -19,10 +19,6 @@ let port = 8085us
 
 let getInitCounter () : Task<Counter> = task { return 42 }
 
-let browserRouter = scope {
```
mexx: It's still used, when `Remoting` is requested

isaacabraham: Doh. Good spot - will fix.

theimowski:
Thanks! looks like the code could be further simplified, i.e. `mainRouter` now has a single `forward` -> could this be inlined?

isaacabraham:
@theimowski yes, it could. I'm actually thinking that I was a little hasty with this - testing it out in Azure App Service it seems not to work!

isaacabraham:
However I think that's an issue here on my branch. I'll confirm shortly.

isaacabraham:
So there's a change needed in Saturn really that I've submitted a PR for. I can make a patch into the template to work around this temporarily.

PR #71, opened 2018-05-01
Azure app insights

isaacabraham:
This modifies the template to include support for App Insights for both Giraffe/Saturn and Suave. Limitations do exist for the Suave implementation e.g. doesn't yet support App Insights live stream, and trace logging isn't hooked up, but the main request / dependency graphs should all work.


isaacabraham:
I think that this is good to go, but needs confirmation that it works not just on my machine :-)

theimowski:
wow this is pretty cool! thanks!

PR #72, opened 2018-05-02
Fixes filepath for Unix-based systems

jeremyabbott:
Caught this while looking at @isaacabraham's Application Insights PR. Verified locally. Works on my machine. ;)


isaacabraham:
Thank you :-)

PR #74, opened 2018-05-03
Add release scripts to gitignore

forki:
@theimowski this is rather urgent since I want to describe release scripts in the docs


theimowski:
Thanks, released in 0.17

PR #77, opened 2018-05-03
Add support for Azure storage

isaacabraham:
This is probably as far as we should go in terms of a "generic" Azure SAFE implementation. It adds a Storage account to the ARM template, links it to the application through a connection string, adds the basic Azure Storage SDK to the Server project and creates an instance of the CloudStorageAccount.

Note: The local development environment (currently) only exists on Windows (and even that requires an installation). Nonetheless, this is still a good starting point for working with Azure Storage (provided we document it!).

### DIFF
--- a/Content/arm-template.json
+++ b/Content/arm-template.json
@@ -10,6 +10,7 @@
         "prefix": "[concat('safe-', parameters('environment'))]",
         "appServicePlan": "[concat(variables('prefix'), '-web-host')]",
         "web": "[concat(variables('prefix'), '-web')]",
+        "storage": "[concat('safe', parameters('environment'), 'storage')]",
         "insights": "[concat(variables('prefix'), '-insights')]"
     },
     "resources": [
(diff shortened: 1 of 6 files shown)


theimowski:
Should this always come with `Azure` option? I.e. what if one wants to use Azure integration without Azure storage?

isaacabraham:
@theimowski I'm only speaking from personal experience here but I'm yet to work on an Azure-ready application that didn't use a Storage account in some way, shape or form.

* The storage account is actually free to create - you only pay when you start to store data in it (and even then it's pennies) so from a "hidden cost" point of view I would say that it's not a problem.
* It's not invasive in terms of the application - you get a CloudStorageAccount object that you can use, or not.

We could make it configurable but even then I would rather have it turned on by default.

theimowski:
fair enough, let me just test all server options and then add it

Content/arm-template.json
```diff
@@ -7,9 +7,11 @@
         "pricingTier": { "type": "string" }
     },
     "variables": {
-        "prefix": "[concat('safe-', parameters('environment'))]",
+        "environment": "[toLower(parameters('environment'))]",
+        "prefix": "[concat('safe-', variables('environment'))]",
         "appServicePlan": "[concat(variables('prefix'), '-web-host')]",
         "web": "[concat(variables('prefix'), '-web')]",
+        "storage": "[concat('safe', variables('environment'), 'storage')]",
```
theimowski: why don't use same naming convention (add dashes) as for other resources?

isaacabraham: Storage accounts have stricter requirements. No dashes, max of 24 characters and all lower case. In fact I'll probably add some more code into the template or script to enforce them all.

PR #81, opened 2018-05-04
Enable AppInsights only if Deploy = 'azure' on Suave server

Zaid-Ajaj:
Running `dotnet new SAFE --Server suave` adds app insights web part to main Suave web part which doesn't work because that only works if the `Deploy = "azure"`. This PR fixes it. 

Also, a couple of names: `init => webApi` and `webPart => webApp` because "webApi is part of webApp" makes a bit more sense than "init is part of webPart" 

Tested locally, but messed up my branch so I just copy pasted it here ;)


isaacabraham:
Argh! Good spot. Thanks for doing this.

PR #89, opened 2018-05-08
Don't use Fulma prerelease

forki:
/cc @MangelMaxime


MangelMaxime:
Do not merge yet, change are needed in the code. I am on it

PR #90, opened 2018-05-08
[WIP] Update to latest Fulma

MangelMaxime:
⚠️ I needed to remove the following line from `paket.dependencies` to make the template work.

```
//#if (Deploy == "azure")
  nuget Microsoft.Azure.Management.ResourceManager.Fluent 1.4.1
  github CompositionalIT/fshelpers src/FsHelpers/ArmHelper/ArmHelper.fs
//#endif
```

Otherwise, the template engine was complaining about `srcOffset` being invalid. Please test in your PR before publishing a new version.

### DIFF
--- a/Content/src/Client/Client.fs
+++ b/Content/src/Client/Client.fs
@@ -11,17 +11,12 @@ open Shared
 
 #if (Fulma != "none")
 open Fulma
-open Fulma.Layouts
-open Fulma.Elements
-open Fulma.Elements.Form
-open Fulma.Components
-open Fulma.BulmaClasses
+open Fulma.BulmaClasses.Bulma.Properties
 #endif
 
 #if (Fulma == "admin" || Fulma == "cover" || Fulma == "hero" || Fulma == "landing" || Fulma == "login")
-open Fulma.BulmaClasses.Bulma
 open Fulma.BulmaClasses.Bulma.Properties
-open Fulma.Extra.FontAwesome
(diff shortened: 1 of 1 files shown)


theimowski:
Strange, which version of `dotnet` you're using?

/cc @forki - this is to be merged to original #89

MangelMaxime:
I am using `dotnet 2.1.103`

MangelMaxime:
Can you please wait like 1-2 days before merging this PR.

I am working on the latest version of Fulma which will support Bulma 0.7.1 (latest) and I am introducing a minor breaking change in the colunm system. Also, I am adding support for modifiers like text alignement etc. 

So can make sense IMO to delay the template update to benefit of this changes.

*The modifier addition is the last thing missing to have 100% of Bulma typed*

isaacabraham:
Can I assume that that issue with the azure dependency is fixed? That can't be removed otherwise it'll break the deployment.

MangelMaxime:
@isaacabraham I didn't tested it since I create the PR but I don't why it would work now on my machine. I didn't change anything.

I will simply update the Fulma part (later in the day) and let you guys look at the issue with `paket.dependencies` if you can repro it or not.

MangelMaxime:
@theimowski This is ready to be merge into #89 

I updated to latest version of Fulma and also removed all the usage of classes string in favor of Fulma API.

theimowski:
awesome! thanks!

PR #95, opened 2018-05-23
Remove configure app from Saturn.

isaacabraham:
Now that Saturn 0.5.0 has been released with built-in default static pages, we can remove it from the SAFE template.


theimowski:
great, thanks!

PR #115, opened 2018-07-09
Solidified Bundle target

psfinaki:
I created the app in the dir: _C:\Users\peter\Desktop\dev life\SAFE_.
Bundle target produces an unclear exception and MSBuild error: "_Only one project can be specified_".

Escaping strings fixes this.


theimowski:
Thanks! Released in 0.24.1

Issue #79, opened 2018-05-03
Fake target `InstallDotNetCore`


MangelMaxime:
Do we really need the `InstallDotNetCore` target ?

If the user want to use the template they need to have `dotnet` already installed. I am asking because, when contributing to a new project this is really painful to have this target. 

I always comment it so it doesn't download dotnet each time I switch from one project to another. 

But perhaps there is good reason for it.

isaacabraham:
I think the only way it can be useful is if it upgrades the local machine when we upgrade the template to the latest SDK. But for clean machines there's no point is there?.

MangelMaxime:
Upgrading local machine can be seen as dangerous I think. It will not impact only the template but all the project on the user system.

isaacabraham:
Agreed. I'm happy to remove that task.

forki:
I vote to keep it. It's very useful for the rest of the team and it upgrades the stuff for everyone if you touch global.json

isaacabraham:
I think that's a separate concern though. I've been burned by this - what if you run a project with an old project.json and it overwrites what you already have? Alternatively, you may not want to upgrade - this happens without your evening knowing.

isaacabraham:
I think that's a separate concern though. I've been burned by this - what if you run a project with an old project.json and it overwrites what you already have? Alternatively, you may not want to upgrade - this happens without your evening knowing.

forki:
not project.json - global.json 

and that file is .NET Core best practice

isaacabraham:
Yeah, sorry - typo. I know they're different :-) Still not sure that's really what we want but open to being educated :-)

MangelMaxime:
If you think this is the good way I am ok with that :)

It's just take forever (5 to 15minutes) to do the first build. And I suppose my issue is more because I work on a lot of project and it's annoying when each one download his own version of dotnet by erasing the previous ones.

What I am trying to say, his perhaps this not an issue for SAFE-Template because in general people with want to use it for their work and not some OSS project.

forki:
It won't take forever if you already have the correct version.

MangelMaxime:
Perhaps, we could simply make it an option in the template ?

forki:
sure that would work

isaacabraham:
And turn it off by default. Especially if you have multiple projects this will just end up ping-ponging between versions.

theimowski:
If we make it an option, no one is ever going to use it I'm afraid. 
[...]

PR #118, opened 2018-07-15
Update remoting to latest, docs for client

Zaid-Ajaj:
This PR updates remoting to latest version and api, I also took the liberty and updated a couple of things along the way, PLEASE REVIEW (I haven't tested every possible combination): 

## Client
Added docs to model, msg, init and update 

`type Model = Counter option` became `type Model = { Counter: Counter option }`  

The messages now became:
```fs
type Msg =
| Increment
| Decrement
| LoadInitialCount 
| InitialCountLoaded of Result<Counter, exn>
```
where `LoadInitialCount` is a message the initiates the request and `InitialCountLoaded` is the one that results from the (http) request, and `init()` becomes:
```fs
// defines the initial state and initial command (= side-effect) of the application
let init () : Model * Cmd<Msg> =
    let initialModel = { Counter = None }
    let initialCmd = Cmd.ofMsg LoadInitialCount
    initialModel, initialCmd
```
## Saturn
`use_router` instead of `router` as per the depreciation message 
`disable_diagnostics` doesn't seem necessary anymore with remoting enabled (probably because the [routing with remoting](https://github.com/Zaid-Ajaj/Fable.Remoting/blob/master/Fable.Remoting.Giraffe/FableGiraffeAdapter.fs#L73) doesn't rely on existing combinators any more) 
## Suave
opening `Suave.Filters` and `Suave.Successful` instead of fully qualifying them with the `OK` and `path` web parts (it is common that they are fully qualified?) 

## build.fsx
`DotNet.Release_2_1_300` => `DotNet.Versions.Release_2_1_300` as per the depreciation message 
`Run` target is `Clean`-ed before running

### DIFF
--- a/Content/build.fsx
+++ b/Content/build.fsx
@@ -42,7 +42,7 @@ let npmTool = platformTool "npm" "npm.cmd"
 let yarnTool = platformTool "yarn" "yarn.cmd"
 //#endif
 
-let install = lazy DotNet.install DotNet.Release_2_1_300
+let install = lazy DotNet.install DotNet.Versions.Release_2_1_300
 
 let inline withWorkDir wd =
     DotNet.Options.lift install.Value
@@ -221,7 +221,8 @@ open Fake.Core.TargetOperators
     ==> "AppService"
 //#endif
 
-"InstallClient"
+"Clean"
+    ==> "InstallClient"
     ==> "RestoreServer"
     ==> "Run"
 
--- a/Content/paket.dependencies
+++ b/Content/paket.dependencies
@@ -15,11 +15,9 @@ group Server
 //#if (!remoting && server != "suave")
     nuget Fable.JsonConverter
 //#elseif (remoting && server == "suave")
-    nuget Fable.Remoting.Server ~> 3.6
-    nuget Fable.Remoting.Suave ~> 2.7
+    nuget Fable.Remoting.Suave 
 //#elseif (remoting && server != "suave")
-    nuget Fable.Remoting.Server ~> 3.6
-    nuget Fable.Remoting.Giraffe ~> 2.7
+    nuget Fable.Remoting.Giraffe 
 //#endif
 //#if (deploy == "azure")
     nuget Microsoft.ApplicationInsights.AspNetCore ~> 2.2
@@ -35,7 +33,7 @@ group Client
     nuget Fable.Elmish.React
     nuget Fable.Elmish.HMR
 //#if (remoting)
-    nuget Fable.Remoting.Client ~> 2.5.1
+    nuget Fable.Remoting.Client
 //#endif
 //#if (layout != "none")
     nuget Fulma
--- a/Content/src/Client/Client.fs
+++ b/Content/src/Client/Client.fs
@@ -17,12 +17,19 @@ open Fulma
 open Fulma.FontAwesome
 #endif
 
-type Model = Counter option
+// The model holds data that you want to keep track of while the application is running
+// in this case, we are keeping track of a counter
+// we mark it as optional, because initially it will not be available from the client
+// the initial value will be requested from server
+type Model = { Counter: Counter option }
 
+// The Msg type defines what events/actions can occur while the application is running
+// the state of the application changes *only* in reaction to these events   
 type Msg =
 | Increment
 | Decrement
-| Init of Result<Counter, exn>
+| LoadInitialCount 
+| InitialCountLoaded of Result<Counter, exn>
 
 
 #if (remoting)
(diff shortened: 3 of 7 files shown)


isaacabraham:
@Zaid-Ajaj indeed, I had the same thoughts. I'm not sure how easy it would be to automate, but yes. Could you create an issue for it?

isaacabraham:
@Zaid-Ajaj there are a few shortcuts around this, however. For example, you can install a template "over the top" of an existing folder, and it usually works fine. This removes the need to do the yarn download etc. etc. Also, did you test out giraffe?

Zaid-Ajaj:
> Could you create an issue for it? 

Yeah, sure

> there are a few shortcuts around this, however. For example, you can install a template "over the top" of an existing folder, and it usually works fine

I will try it out 

> Also, did you test out giraffe?

No, not yet, will do sometime today :)

Krzysztof-Cieslak:
> disable_diagnostics doesn't seem necessary anymore with remoting enabled
I've fixed that in latest - `site.map` generation doesn't crash anymore if it hasn't detected any routes.

theimowski:
@Zaid-Ajaj see https://github.com/SAFE-Stack/SAFE-template/commit/ad8bcbe44b36dd419a7835193a2c8e60af2d477d - I've added a FAKE target to help with the test workflow.
Not very much, but now you can `fake build -t install` and have the updated template already available in `dotnet`

Zaid-Ajaj:
@theimowski Thank you! I will test it out next time :smile: 

I have now tested giraffe with and without remoting and it seems to work (On My Machine™), I am working on windows 10 with dotnet 2.1.300 

I came across another issue, but I am noticing `global.json` is magically coming into existence and disappearing on it's own after I execute the `Run` for the second time in a different directory, having the effect of the `Run` target not running the server and I get:
```
(The process cannot access the file 'C:\projects\safe-giraffe\no-remoting\global.json' because it is being used by another process.)
```` 
only after I exit the process. This seems due to this block:
```fs
let install = lazy DotNet.install DotNet.Versions.Release_2_1_300

let inline withWorkDir wd =
    DotNet.Options.lift install.Value
    >> DotNet.Options.withWorkingDirectory wd
```
my workaround is to replace `runDotNet` with `runTool "dotnet" cmd workingDir`

theimowski:
I'm considering removing the `DotNet.install` alltogether - we have .NET SDK 2.1 in the prerequisites anyway

Relates to https://github.com/SAFE-Stack/SAFE-template/issues/79

Zaid-Ajaj:
This can be merged, if I missed something, just let know :)

theimowski:
Actually there are 2 things (posted comments in code diff):
1. How about float pinning the `Fable.Remoting` dependencies? 
e.g. `nuget Fable.Remoting.Suave ~> 3` to prevent future breaking changes?
2. I'm not convinced with adding one more Msg case - IMO it makes the `update` function more complex than it needs to be

I'm fine with rest of changes

Zaid-Ajaj:
> How about float pinning the Fable.Remoting dependencies?

That shouldn't be needed anymore when pulling lastest packages (also, I will try to publish one library with it's dependencies in one package) 

> I'm not convinced with adding one more Msg case - IMO it makes the update function more complex than it needs to be

In the context of this simple counter app, I agree with you. 

The reasoning behind this is for larger apps: usually a component doesn't ask for it's data when initialized at application startup, but when loaded or navigated to from other components where you will explicitly need to dispatch a message to initiate the request (hence the `LoadInitialCounter`) 

I can remove the message for now, but since we are on the subject, I was actually thinking of using a more representative app in the template like a to-do list instead of a counter: it showcases the used frameworks much better with a CRUD scenario while still being [very simple](https://github.com/Zaid-Ajaj/SAFE-TodoList/blob/master/src/Client/State.fs) to follow and understand 

What do you think? @isaacabraham @Krzysztof-Cieslak

theimowski:
I'm happy to discuss the idea or more representative app, but let's make it a separate issue

isaacabraham:
I'm definitely off the view that the template should be as simple as possible in terms of code. I really don't want to see it include CRUD stuff that is opinionated in any way or might not be applicable - we want the template to be useful for any SAFE app really.

I'd say there's a good case to be made for creating such an example app to include in the SAFE github area like the bookstore etc

Zaid-Ajaj:
I see, then we will keep it simple :smile:

theimowski:
It failed on:
* build script with change `DotNet.Versions` (locked Fake.DotNet.Cli didn't have that module
* saturn `application` CE - I got compiler error on `url` with Saturn pinned to 0.5

theimowski:
I've rebased the commits now on new master, the changes are released in 0.26 - thanks again for contribution!

Issue #136, opened 2018-08-03
Support Heroku


Nhowka:
I made a [test project](https://github.com/Nhowka/BridgeChat) based on the current template and modified [a buildpack](https://github.com/Nhowka/dotnetcore-buildpack) to support the fake script on it. For the deploy button to work I only needed an `app.json` defining the buildpack.

Should we have that under the SAFE-Stack umbrella and an option to generate that `app.json` file on the template?

Zaid-Ajaj:
Yes, please! 🙏 I specialized "SAFE buildpack" would be awesome

isaacabraham:
No objection here! Would you also consider documenting it in the safe docs repo? thanks

Nhowka:
Ok by me! Maybe the [Procfile](https://github.com/Nhowka/dotnetcore-buildpack/blob/master/bin/compile#L49-L51) generation should also be on the script to support different project names. I'll work on it and ask for help as needed.

theimowski:
closed by #251

PR #137, opened 2018-08-04
Add Heroku support

Nhowka:
Implements #136 

I'm using `https://github.com/SAFE-Stack/SAFE-buildpack` as a placeholder. I created a [branch](https://github.com/Nhowka/dotnetcore-buildpack/tree/SAFE) on my modified buildpack for supporting it. You can change it to `https://github.com/Nhowka/dotnetcore-buildpack#SAFE` if you don't want to have it under the org name.

I also had to change the `--port free` position because it wasn't being recognized. Weird.

### DIFF
--- a/Content/.template.config/template.json
+++ b/Content/.template.config/template.json
@@ -109,6 +109,10 @@
                 {
                     "choice": "docker",
                     "description": "additional FAKE targets to bundle and build Docker image"
+                },
+                {
+                    "choice": "heroku",
+                    "description": "additional FAKE targets to deploy on Heroku"
                 }
             ]
         }
@@ -203,6 +207,10 @@
                 {
                     "exclude": "arm-template.json",
                     "condition": "(deploy != \"azure\")"
+                },
+                {
+                    "exclude": "app.json",
+                    "condition": "(deploy != \"heroku\")"
                 },
                 {
                     "exclude": "yarn.lock",
--- a/Content/app.json
+++ b/Content/app.json
@@ -1,0 +1,9 @@
+{
+    "buildpacks": [
+    {
+      "url" : "heroku/nodejs"
+    },
+    {
+      "url" : "https://github.com/SAFE-Stack/SAFE-buildpack"
+    }]
+  }
--- a/Content/build.fsx
+++ b/Content/build.fsx
@@ -98,7 +98,7 @@ Target.create "RestoreServer" (fun _ ->
 
 Target.create "Build" (fun _ ->
     runDotNet "build" serverPath
-    runDotNet "fable --port free webpack -- -p" clientPath
+    runDotNet "fable webpack --port free -- -p" clientPath
 )
 
 Target.create "Run" (fun _ ->
@@ -106,7 +106,7 @@ Target.create "Run" (fun _ ->
         runDotNet "watch run" serverPath
     }
     let client = async {
-        runDotNet "fable --port free webpack-dev-server" clientPath
+        runDotNet "fable webpack-dev-server --port free" clientPath
     }
     let browser = async {
         do! Async.Sleep 5000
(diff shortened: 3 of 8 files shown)


Zaid-Ajaj:
Will this not install `dotnet` twice? once by the build pack to be able to globally install `fake-cli` and once again from inside `build.fsx`?

Nhowka:
Yes. Using specific version it failed, but didn't when set to the default.

Help me! How do I use the existing version?

Zaid-Ajaj:
Which specific version did you use and why did it fail? if you choose version 2.1.300 from the build pack, it should be good enough to install the fake-cli and you can delete the dotnet (re)installation from the build.fsx, this operation of installing dotnet from build.fsx was already giving lots of headaches for local dev as it creates and locks a global.json file after restarting the app

Nhowka:
It fails because of dependencies. If I remember correctly it was `libunwind`. I tried some stuff and that clearly shows that I don't know what I'm doing.

Nhowka:
It's probably ready, now. I created [this repo](https://github.com/Nhowka/safe-buildpack) with a cleaner version of the buildpack. I can transfer the ownership of it.

Nhowka:
Nope, other test failed on my fork. The tests are being very random...

theimowski:
The tests are random on purpose - the number of all possible combinations for template options is too big, so I'm using FsCheck generator to choose 10 various set of options, they'll be different each time. If 10 is not big enough we can increase it

Nhowka:
With those files it is just pushing to the Heroku git, there is a guide here: https://devcenter.heroku.com/articles/git

I will transfer the ownership of the buildpack now.

theimowski:
Thanks, I'll test that in coming days

Nhowka:
I checked and it's creating the `Procfile` with the wrong base, I'll try to fix it.

theimowski:
I also got a request from GitHub:

`@Nhowka wants to transfer the Nhowka/SAFE-buildpack repository to theimowski/SAFE-buildpack`

I think you wanted to move to `SAFE-Stack/SAFE-buildpack` instead? 
Is it ok to add that repository to org? @SAFE-Stack/sample-app-admins

Nhowka:
Yes, I couldn't transfer to the org. Feels like the bundle target didn't created the deploy folder, probably the Procfile should be in the ignore list too. It's like the buildpack never ran.

theimowski:
I think that `bundle` target creates the `deploy` dir, but it's gitignored - should it be pushed to heroku remote?

Nhowka:
Yes, that means that the buildpack didn't work. Did you changed the path on `app.json` to the real repository?

theimowski:
yes, changed it to https://github.com/Nhowka/SAFE-buildpack

Nhowka:
So probably it won't add a new buildpack to an already created project, just to projects that are being created from the repository. I'll add a new target that configures the buildpack into the current repository.

theimowski:
@Nhowka do you have any update to this?

Nhowka:
My main machine stopped working, I'll try to revisit that on the weekend on another machine. The progress I had before that was about adding some confirmation of the existence of the heroku cli and checking if the folder already was part of a heroku project or not. Probably a new target just for initing the project and adding the buildpack will need to run once before being able to deploy.

andrzejsliwa:
@Nhowka any progress on it?

Nhowka:
I got some free time and a working machine. I did some tests updating the buildpack and got another heroku-based project to start fine without needing mono using paket as a global tool. That simplifies a lot.

 I'll keep this PR open to use as reference, but probably I'll need to create a new PR with a new branch. Unless some git guru show me the right way to bring these changes to the current version with less work.  😋

Nhowka:
Use #251.

Issue #66, opened 2018-04-26
Include paket.lock file?


isaacabraham:
Given that we're now including a `yarn.lock` file, shouldn't we do the same with `paket.lock`? Not only will this increase the speed for the first build (important IMHO), but will also allow us to ensure stable releases.

Krzysztof-Cieslak:
Uhh, we don’t do that? Yeah, we definitely should, IMO

theimowski:
Yep I think the advantage of fast first build doesn't outgrow the cost of having unused dependencies in the project. Closing for now

theimowski:
reopening - it's getting more and more irritating, specially since stuff like https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/965 happens and one cannot have deterministic build when creating from given version of template...
@matthid suggested using groups for this, that could work but would require quite a lot of effort

supertom44:
I just came to create a new project from the template and got caught out by the above-linked error.

Is there a quick workaround while the issue is being resolved?

Cheers.

isaacabraham:
@theimowski how about a compromise here - let's either create a paket lock for when the default options only are selected. Otherwise, it doesn't create one. OR - we create a massive lock file that contains everything for all packages (this potentially introduces other problems though).

I demoed creating a SAFE app from scratch the other night and had to pre-generate the lock file because it simply took too long otherwise.

forki:
Yes lock file for default seems a good solution

Isaac Abraham <notifications@github.com> schrieb am So., 24. Juni 2018,
17:36:

isaacabraham:
I've just tested out with a clean safe template and included a ready-made paket.lock, it definitely helps a lot. The biggest slow downs that I saw now are (I think) compilation of the javascript side and also this "linking" process of the yarn dependencies. Is there any way that we can "precompute" that as well?

forki:
Yes if we add a paket.lock then we should add yarn's lock file as well.

Isaac Abraham <notifications@github.com> schrieb am So., 24. Juni 2018,
23:33:

isaacabraham:
The lock file is already there.

isaacabraham:
@forki this is what I'm referring to https://github.com/yarnpkg/yarn/issues/1496

forki:
I don't think I have this issue.

Isaac Abraham <notifications@github.com> schrieb am Mo., 25. Juni 2018,
23:16:

[...]

PR #139, opened 2018-08-06
Generate paket.lock for all possible template options

theimowski:
The idea here is to write couple of Fake helper targets for dealing with various paket.lock files in the template.
Because the paket.lock will always have 3 groups:
* Build
* Client
* Server

we can treat each of the group separately to minimize total number of paket.lock permutations.

Workflow in this PR for working with paket.lock

* `GenPaketLockFiles` Fake target can generate lock files for representative number of template options, split it for each group and save in `./Content/{Group}` directory
* `BuildPaketLockFiles` can take the lock files from corresponding `./Content/{Group}` dirs and build all permutations of lock files in `./Content`. This target is now dependency of `Pack` target
* `RemovePaketLockFiles` occurs after `Pack` target so that the git repo is not littered with generated paket.lock files
* `GenJsonDefinitions` target can be invoked to print conditions needed to be placed into `template.json` manifest
* `UpdatePaketLockFiles` updates all paket lock files (can pass in specific group name(s) as target parameter)

Refers to #66


isaacabraham:
Looks very interesting! Will this generate the lock files in advance or when people create a new project from the template?

theimowski:
it creates the lock files before packaging up the NuGet, so they are present in the template already

isaacabraham:
How many are there currently?

theimowski:
```
type ServerPaketDependencies =
    { Server : ServerPaketDependency
      Remoting : bool
      Azure : bool }

type CombinedPaketDependencies =
    { Azure : bool
      Remoting : bool
      Fulma : bool
      Server : ServerPaketDependency }
```

3 x 2 x 2 x 2 = **24**

Issue #110, opened 2018-06-27
Change `remoting` parameter to allow adding new options


theimowski:
Related to #107 

I'd vote for changing the boolean `--remoting` parameter to choice parameter with a catchy name (how about `--link` ?) and `remoting` as the only option for now.

So instead of `dotnet new SAFE --remoting` one would type `dotnet new SAFE --link remoting`

This will allow adding easily support for Elmish.Bridge in future, without "breaking" CLI invocation

isaacabraham:
Will there be the option to add both bridge and remoting together?

Nhowka:
Could be. The main problem is having the room without changing the counter. Maybe adding a new feature only when bridge is used with remoting could be an option.

theimowski:
Closed by #143

PR #143, opened 2018-08-16
Change 'remoting' to 'communication' - references #110

theimowski:


Nhowka:
Does that enables multiple communications choices?

theimowski:
For now it's just renaming the parameter, it should make a bit easier to add new choices in future

PR #146, opened 2018-08-29
Move .NET Sdk version to global.json

0x53A:
I've kept the version at 2.1.300, but I think it should be updated to 2.1.401, especially now that that version is LTS.


theimowski:
#79 #111

0x53A:
Ah, I think I understand why you moved it from global.json to the script.

If the version in global.json isn't already installed, then it won't even execute the script, making the Target kinda useless...

theimowski:
Not really, I moved sdk version from `global.json` to build.fsx when upgrading to FAKE 5 - I copied that from FAKE repository itself, assuming it will become a new standard for projects using FAKE 5.

We make a prerequisite on FAKE 5 installed as a global tool, so even if version specified in global.json is not already installed, it should work

theimowski:
I've decided to remove "Install dotnet" step for now - https://github.com/SAFE-Stack/SAFE-template/commit/e4462c98116e9b93dbf66aefaf1335a7eca853ea 

Let's see what people think of it - if you encounter issues please reopen

PR #157, opened 2018-09-27
Upgrade to Fable 2

theimowski:
#144


MangelMaxime:
Hello @theimowski you need to update your webpack config and some npm dependings too.

For the webpack config, you can based yourself on:
- [Coming from fable repo](https://github.com/fable-compiler/webpack-config-template/blob/master/webpack.config.js)
- [Coming from my previous work on booktstore](https://github.com/SAFE-Stack/SAFE-BookStore/pull/349)
- You need to use the beta version of `fable-loader` seems like we didn't released the stable version yet
- Use `@babel` scope for including babel your project. [See this commit](https://github.com/MangelMaxime/fulma-demo/commit/4b0a943f42ba24d47a4ac5cc567b06b4c5850719)

MangelMaxime:
I think with this step you should be ready or close to be. If news errors I coming please tell me I will help you or send a PR to your PR when I have time.

MangelMaxime:
@theimowski Your welcome

If this can help you here is a WIP guide for migration. In theory, all the steps are in but it is not well written yet :) And if needed ping me here or gitter/slack.

theimowski:
@MangelMaxime I think I've migrated most of the configuration - can you have a look?
I didn't however add the [optimization](https://github.com/MangelMaxime/fable-suave-scaffold/blob/33efc0fab24c8785e244c5966c54fce2ba03e65a/src/Client/webpack.config.js#L66-L77) to webpack - webpack-dev-server didn't work for me, not sure why

@Zaid-Ajaj with Fable.Remoting I'm getting a lot of warnings like below for Fable.Persimmon, do you know why?

```
WARNING in ./.fable/Fable.Parsimmon.3.0.0/Parsimmon.fs 83:31-34
"export 'all' was not found in './Parsimmon.js'
 @ ./.fable/Fable.SimpleJson.2.3.0/SimpleJson.fs
 @ ./.fable/Fable.Remoting.Client.4.1.0/Proxy.fs
 @ ./.fable/Fable.Remoting.Client.4.1.0/Remoting.fs
 @ ./src/Client/Client.fs
 @ ./src/Client/Client.fsproj
 @ multi (webpack)-dev-server/client?http://localhost:8080 (webpack)/hot/dev-server.js whatwg-fetch @babel/polyfill ./src/Client/Client.fsproj
```

MangelMaxime:
@theimowski You need to exclude `.fable` folder from the template files as you do for `.fake`.

Otherwise, the build fail until your delete the`.fable` folder and re-run the build.

MangelMaxime:
I will try to take a look at your PR tomorrow.

Zaid-Ajaj:
@theimowski These warnings are fixed in the latest versions of Fable.Parsimmon as I don't get them anymore in either Fable.Remoting nor Fable.SimpleJson (dependants of Fable.Parsimmon), please try deleting Fable.Parsimmon from the .fable cache as well as the .nuget cache on your local machine

MangelMaxime:
@theimowski If I understand correctly, `webpack-dev-server` is working correctly with current status of the PR ? *This is what I see when trying it locally*

But when you add, the `optimization` then it's failing with a white page ? *This is the behavior I got*

In order to fix that, you need to add `<script src="./js/vendors.bundle.js"></script>` to your `index.html` before loading `bundle.js`. 

The optimization split your code in `2` files:
- `vendors.bundle.js`, which contains all the code coming from `node_modules` like react, whatwg-fetch, etc.
- `bundle.js`, which contains the user code and the one coming from Fable

This is why in general I use [html-webpack-plugin](https://github.com/jantimon/html-webpack-plugin) so it inject automatically the needed entry to your `index.html`. Especially useful, when you generated `hased` name for your files so the browser do not cache them.

theimowski:
@Zaid-Ajaj I've removed Fable.Parsimmon from nuget cache, and tried to create a project from this PR with `dotnet new --communication remoting` - unfortunately I'm still getting the warnings, although I have only Fable.Parsimmon 3.0.0 in both `./.fable/` and `~/.nuget/packages/`

theimowski:
Thanks @MangelMaxime - yes that was exactly my issue, after adding the `vendors.bundle.js` it started working properly.

We can think of adding html webpack plugin later

theimowski:
I'm not sure why but it seems that after adding the `optimization` to webpack, for production it generates `bundle.js` and `1.bundle.js`. Where is this configured and what's the difference between debug and production ?

MangelMaxime:
@theimowski Hum this is strange, in general `webpack` generate the same modules for dev and production mode... I don't know. Do you have the `vendors.bundle.js` generated in production too ? Or is it replaced by `1.bundle.js` ?

theimowski:
For production (after `fake build`) there's:
* 1.bundle.js
* bundle.js

and for dev (after `fake build -t run`) I can't see all artifacts but webpack can serve:
* vendors.bundle.js
* bundle.js

MangelMaxime:
Ok, I will take a look at it.

Zaid-Ajaj:
@theimowski Hmm weird, I will investigate this, I will let you know how it goes

theimowski:
@MangelMaxime Thanks - this seems to do the job. Now it generates constantly `vendors.js` and `app.js` both for dev and prod

Content/build.fsx
```diff
@@ -92,15 +92,15 @@ Target.create "RestoreServer" (fun _ ->
 
 Target.create "Build" (fun _ ->
     runDotNet "build" serverPath
-    runDotNet "fable webpack-cli -- --config src/Client/webpack.config.js -p" clientPath
+    runDotNet "fable webpack-cli --port free -- --config src/Client/webpack.config.js -p" clientPath
```
MangelMaxime: You can remove `--port free` it's now the default for Fable 2.

Content/build.fsx
```diff
@@ -92,15 +92,15 @@ Target.create "RestoreServer" (fun _ ->
...
 )
 
 Target.create "Run" (fun _ ->
     let server = async {
         runDotNet "watch run" serverPath
     }
     let client = async {
-        runDotNet "fable webpack-dev-server -- --config src/Client/webpack.config.js" clientPath
+        runDotNet "fable webpack-dev-server --port free -- --config src/Client/webpack.config.js" clientPath
```
MangelMaxime: You can remove `--port free` it's now the default for Fable 2.

theimowski:
I have hard times trying to figure out what's the matter with Fable.Parsimmon - right now it's the only remaining issue for this PR, thought it might be related to webpack.config being in a different directory than package.json, but that's not the case

Zaid-Ajaj:
@theimowski I haven't gotten around to investigate this thoroughly but it must be a config thing because I use Parsimmon both in SimpleJson and Remoting.Client and I don't get these warnings, for example this is [the build log](https://travis-ci.org/Zaid-Ajaj/Fable.SimpleJson/builds/4373052640) for SimpleJson from travis ci, it has no warnings and [this is my config](https://github.com/Zaid-Ajaj/Fable.SimpleJson/blob/master/webpack.config.js), maybe it has something to do with [babel config in these lines](https://github.com/Zaid-Ajaj/Fable.SimpleJson/blob/master/webpack.config.js#L3-L10)?

theimowski:
@Zaid-Ajaj from what I can see Fable.SimpleJson is still using parts of [Fable 1 stack](https://github.com/Zaid-Ajaj/Fable.SimpleJson/blob/master/package.json#L14), whereas for SAFE I'm trying to follow Maxime's [changes](https://github.com/SAFE-Stack/SAFE-BookStore/pull/349/files) in package.json with [those dependencies](https://github.com/SAFE-Stack/SAFE-template/blob/b85d184cac45d95965191c3272b5f14d9ece77d8/Content/package.json)

I'm still learning the fable compilation phases and included dependencies, but I'm wondering whether the new tools as recommended [here](http://fable.io/blog/Migration-to-Fable2.html) would actually build Fable.SimpleJson successfully.

Zaid-Ajaj:
@theimowski Both simplejson and parsimmon are succesfully compiled against Fable 2, otherwise the code wouldn't work as it uses Fable 2 specific constructs (type metadata injection) but it seems that because the test projects are simple enough, they didn't need require "full fable 2 treatment" and they run with the `env` preset of babel instead of the new `@babel/preset-env`, 

> Side note: `fable-util` is forgotten there but is isn't used anywhere 

I have cleaned up everything and made the configuration follow the migration guide and found the problem: the embedded `.js` dependency of Parsimmon exports the module using ES5 syntax (i.e. `module.exports = { ... }`) but either babel-loader or fable-loader isn't happy with that as it expected that I explicitly `export const everyValue  = ... ` in ES6 syntax,  which I did and it solved the problem (cc @alfonsogarciacaro) 

Now all the packages (parsimmon, simple json and remoting client) are patched, please update the remoting client to `4.2.0` and see if the warnings are gone

theimowski:
@Zaid-Ajaj Yes sorry, that was what I meant - Fable dependencies in SimpleJson were different than those used in migration guide.

Thank you very much for taking care of updating the packages - I can confirm no more warning for safe template using fable.remoting in this PR.

This is now ready for final review, if anyone is interested - @MangelMaxime @alfonsogarciacaro @isaacabraham @Zaid-Ajaj

Issue #166, opened 2018-10-29
Add option to include Fable.Reaction


theimowski:
https://github.com/dbrattli/Fable.Reaction

StachuDotNet:
Some context:
There was a quite healthy discussion around Fable.Reaction at FableConf during @dbrattli's Reaction workshop. There are some clear trade-offs to using Fable.Reaction in certain contexts, and including this library as part of the official SAFE template would allow for higher usage and development of ideas.

Perhaps @dbrattli would have some thoughts on what this may look like?
SAFE team, any initial thoughts on this?

theimowski:
I'd like to start with simplest possible thing: add new option to CLI, e.g. `--pattern reaction` which would spin up the same code that was presented during workshops

isaacabraham:
@theprash had some concerns regarding whether it would "break" the MVU model or not - perhaps he can expand upon it here. I've not seen it yet and so can't comment on it, but it sounds like for some cases it would be of great benefit e.g. debouncing and the like.

Also is the "risk" (if that's even the right word) of providing a more complex set of choices in terms of packages and more fundamentally, how people should be writing SAFE apps... e.g. what impact would it have on things like Bridge? Do you "throw away" the MVU pattern completely or have some hybrid approach?

Hopefully this doesn't come across that I'm against this per se (and I need to try it out myself to form an opinion) - I just think we should be aware about the overall impact and effort that might be needed to introduce this into the official stack as such (especially documentation).

dbrattli:
Fable.Reaction is basically just a [combination](https://github.com/dbrattli/Fable.Reaction/blob/master/src/Program.fs#L43) of an Elmish [subscription](https://elmish.github.io/elmish/subscriptions.html) and a composed `dispatch`, so it's nothing special from what you could already do within Elmish. I guess the controversial issue is that we compose side-effects into `dispatch` using `Bind` instead of going the round trip with `update` returned commands. But isn't `dispatch` a side-effect ('a -> unit) in itself? Is there a really a difference between a `dispatch` function called when a button is clicked, and a composed `dispatch` that (suddenly) emits a button click? 

[...]

PR #168, opened 2018-10-30
WIP: New option for Fable.Reaction

theimowski:
#166


isaacabraham:
I assume that this will be a post-v1 addition?

theimowski:
I don't have a plan for that yet - let's see where it goes and decide then. I won't merge it until we agree it's a good idea

theimowski:
I think I'd like to see it before v1 - I've added https://github.com/SAFE-Stack/docs/pull/79 to document how the parameter could look like.

PR #182, opened 2018-11-08
Counter as a record

### LINKED ISSUE #177: Use "full" record in counter sample
Any objection to replacing the use of `int` with a proper record? The use of a type alias, whilst cute, is somewhat confusing to newcomers and also breaks down as soon as you want to add a second field to a record etc.

isaacabraham:
Fixes #177.


Content/src/Client/Client.fs
```diff
@@ -107,11 +107,11 @@ let query msgs =
 // The update function computes the next state of the application based on the current state and the incoming events/messages
 let update (msg : Msg) (currentModel : Model) : Model =
     match currentModel.Counter, msg with
-    | Some x, Increment ->
-        { currentModel with Counter = Some (x + 1) }
-    | Some x, Decrement ->
-        { currentModel with Counter = Some (x - 1) }
-    | _, InitialCountLoaded (Ok initialCount)->
+    | Some counter, Increment ->
```
Nhowka: What if instead of `Some counter` and then using `counter.Value` we use `Some { Value = value }` and return `Some { Value = value + 1 }`?

isaacabraham: That'll work too, although I'm not sure how much cleaner that is. Happy to make that change though.

build.fsx
```diff
@@ -29,7 +29,7 @@ let formattedRN =
 
 Target.create "Clean" (fun _ ->
     Shell.cleanDirs [ nupkgDir ]
-    Git.CommandHelper.directRunGitCommandAndFail "./Content" "clean -fxd"
+    // Git.CommandHelper.directRunGitCommandAndFail "./Content" "clean -fxd"
```
theimowski: Has this been an issue? 
I added that since there used to be a problem where I packaged up a lot of unnecessary files to the nupkg making it bigger than it should be.
I know this can be frustrating when adding a new file to the `./Content` dir, but otherwise it's a safety net for not pushing random stuff into package.

Content/src/Client/Client.fs
theimowski: I'm ok with both approaches, but see comment below

theimowski:
LGTM, just 2 minor comments

build.fsx
isaacabraham: Doh sorry. I didn't mean to commit that line - I'll revert it. But yes, that's why I commented it out - just temporarily. Good spot.

isaacabraham:
@theimowski let me know if that last commit fixed the issue - if so I'll merge

theimowski:
I did also mean to change the `update` function to object-update expression, but I'll merge and correct that myself. 
Thanks!

PR #183, opened 2018-11-08
Add README.md file to SAFE Template content

WalternativE:
As a user of the SAFE template I might want to have an initial Readme which tells me how to run/deploy the newly created application. I don't know how much value it adds as people learning about the template most likely do so from the SAFE documentation site but at least for other contributors or people who just find the repo in source control it could be helpful.

### DIFF
--- a/Content/README.md
+++ b/Content/README.md
@@ -1,0 +1,49 @@
+# SAFE Template
+
+This template can be used to generate a full-stack web application using the [SAFE Stack](https://safe-stack.github.io/).
+
+## Install pre-requisites
+
+You'll need to install the following pre-requisites in order to build SAFE applications
+
+* [FAKE 5](https://fake.build/) installed as a [global tool](https://fake.build/fake-gettingstarted.html#Install-FAKE)
+* The [Yarn](https://yarnpkg.com/lang/en/docs/install/) package manager (you an also use `npm` but the usage of `yarn` is encouraged).
+* [Node LTS](https://nodejs.org/en/download/) installed for the front end components.
+* If you're running on OSX or Linux, you'll also need to install [Mono](https://www.mono-project.com/docs/getting-started/install/).
+
+## Work with the application
+
+To concurrently run the server and the client components in watch mode use the following command:
+
+```bash
+fake build -t Run
+```
+
+//#if (deploy == "docker")
+To build your application as a docker container run the following command:
+
+```bash
+fake build -t Docker
+```
+//#endif
+//#if (deploy == "azure")
+To deploy your application as an Azure Web App run the following command:
+
+```bash
+fake build -t AppService
+```
+//#endif
+
+## SAFE Stack Documentation
+
+You will find more documentation about the used F# components at the following places:
+
+* [Saturn](https://saturnframework.org/docs/)
+* [Fable](https://fable.io/docs/)
+* [Elmish](https://elmish.github.io/elmish/)
+
+If you want to know more about the full Azure Stack and all of it's components (including Azure) visit the official [SAFE documentation](https://safe-stack.github.io/docs/).
+
+## Troubleshooting
+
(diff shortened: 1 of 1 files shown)


StachuDotNet:
Absolutely a big value-add here.

isaacabraham:
@WalternativE first thanks for this :-) I've no objection to this at all in principle.

The only question I have is around duplication of text - we have the quick start [here](https://safe-stack.github.io/docs/quickstart/) and could link to that instead?

Alternatively, perhaps we could include the file from that repository as a paket dependency into this repo and rename it as readme.md? Thoughts?

Krzysztof-Cieslak:
So, I don’t think we should only link to the webpage, having some readme in generated project seems like an good thing.

Coping file may be good idea, on the other hand it seeems like the type of design where you try so hard to abstract over something to not repeat any code that you introduce strong coupling between components ;-)

isaacabraham:
@Krzysztof-Cieslak I just don't want to be in a position where we end up with out of date docs in one of the two repos.

WalternativE:
Referencing the quick start guide was my initial thought (as you may have recognized I 'lifted' the major part of the documentation from it). I adapted it a bit because I reckon that people working with the project don't have to necessarily be the ones who used the template to create it (finding it on GitHub, on the company SCM, etc). Also there are more (or different) options that should be documented in the README that would be too much for the quick start (like the docker or Azure build targets).

I think it would be a good idea to reference the quick start guide right at the top to tell people that the application was created using the template in this quick and easy way. Also I'd add the dependency on the dotnet Core SDK again - I deleted it because I thought it would be obvious that people have it installed as they used the template but as of my own reasoning this doesn't have to be the case 😅

Would this be good or am I missing your point?

isaacabraham:
Yeah - I think in this case, less is more with the readme - and what you're put here is way, *way* better than what we currently have, which is nothing.

Content/README.md
```diff
@@ -0,0 +1,49 @@
+# SAFE Template
+
+This template can be used to generate a full-stack web application using the [SAFE Stack](https://safe-stack.github.io/).
```
isaacabraham: Add a link here to the quick start as well, from which this is based upon?

Content/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+You will find more documentation about the used F# components at the following places:
+
+* [Saturn](https://saturnframework.org/docs/)
+* [Fable](https://fable.io/docs/)
+* [Elmish](https://elmish.github.io/elmish/)
+
+If you want to know more about the full Azure Stack and all of it's components (including Azure) visit the official [SAFE documentation](https://safe-stack.github.io/docs/).
+
+## Troubleshooting
```
isaacabraham: I think you can remove this section - it's something that will be fixed (if it's not already fixed) relatively soon.

Content/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Work with the application
+
+To concurrently run the server and the client components in watch mode use the following command:
+
+```bash
+fake build -t Run
+```
+
+//#if (deploy == "docker")
```
isaacabraham: I would personally remove the docker and app service bits - you need to really read the docs to understand how to use them both. Perhaps replace with a couple of links to those pages, or just remove.

Content/README.md
WalternativE: Do you mean the Troubleshooting paragraph  or the paragraph before that and the Troubleshooting paragraph?

theimowski: I think I saw newest SDK warning about that when installing a global tool, at least on a bare alpine-sdk2.1 docker image

WalternativE: Ok I'll just leave it then for now.

theimowski:
LGTM, @isaacabraham do you have any other comments?

isaacabraham:
LGTM - thanks again :-)

PR #186, opened 2018-11-13
Add more comments in webpack.js

isaacabraham:
Added more comments to the `webpack.config.js` file. I also removed the CONFIG object and embedded it directly into the configuration - I'm not sure if this is better or worse, happy to revert that, but for me as a non-webpack person it means one less object to reason about in the file.

Also renamed the server proxy port to `SERVER_PROXY_PORT ` - otherwise it could be confusing for people using different web servers. Theoretically this is a breaking change so I would like to get this in before v1.

### DIFF
--- a/Content/src/Client/webpack.config.js
+++ b/Content/src/Client/webpack.config.js
@@ -6,55 +6,30 @@ function resolve(filePath) {
     return path.join(__dirname, filePath)
 }
 
-var CONFIG = {
-    fsharpEntry: {
+// passing -p on the command line to webpack specifies production mode, otherwise debug is used.
+var isProduction = process.argv.indexOf("-p") >= 0;
+
+console.log("Bundling for " + (isProduction ? "production" : "development") + "...");
+
+module.exports = {
+    entry: {
         "app": [
             "whatwg-fetch",
             "@babel/polyfill",
             resolve("./Client.fsproj")
         ]
     },
-    devServerProxy: {
-        '/api/*': {
-            target: 'http://localhost:' + (process.env.SUAVE_FABLE_PORT || "8085"),
-            changeOrigin: true
-        }
-    },
-    historyApiFallback: {
-        index: resolve("./index.html")
-    },
-    contentBase: resolve("./public"),
-    // Use babel-preset-env to generate JS compatible with most-used browsers.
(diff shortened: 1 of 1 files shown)


theimowski:
/cc @MangelMaxime 

I grabbed the `CONFIG` object from Maxime's suggestions - I think the goal here was to design the webpack config so that the only place a user needs to configure is this CONFIG thing.

MangelMaxime:
The `CONFIG` object was an idea from Alfonso so people only need to modify this object and don't need to look at the whole `webpack.config.js`.

But I think, if you are using a template you should not need to modify the `webpack.config.js` at first. And when you need to modify it, this means you are going to not follow the default configuration and probably going to use advanced features and so learn/understand webpack.

So it's up to you depending on what you prefer. Personnaly, at work I don't have the `CONFIG` object and only have `commonPlugins` and `babelOptions` as variables in order to not declare twice the same things.

```js
// Common/shared variables

var babelOptions = {
    presets: [
        ["@babel/preset-env", {
            "targets": {
                "browsers": ["last 2 versions"]
            },
            "modules": false,
            "useBuiltIns": "usage"
        }],
        "@babel/react"
    ],
    "plugins": [
        "@babel/plugin-proposal-class-properties",
        "@babel/plugin-proposal-object-rest-spread"
    ]
};

var commonPlugins = [
    new HtmlWebpackPlugin({
        filename: './index.html',
        template: './Sources/index.html'
    }),
];

// Usage

module.exports = {
    // ...
    plugins: isProduction ?
        commonPlugins.concat([
            new MiniCssExtractPlugin({
                filename: 'style.[hash].css'
            })
        ])
        : commonPlugins.concat([
            new webpack.HotModuleReplacementPlugin(),
            new webpack.NamedModulesPlugin()
        ]),
    // ...
    module: {
        rules: [
            {
                test: /\.fs(x|proj)?$/,
                use: {
                    loader: "fable-loader",
                    options: {
                        babel: babelOptions
                    }
                }
            },
            {
                test: /\.js$/,
                exclude: /node_modules/,
                use: {
                    loader: 'babel-loader',
                    options: babelOptions
                },
            },
        ]
    }
    // ...
};
```

theimowski:
Fair enough, let's inline that `CONFIG` object then. Merging - thanks @isaacabraham  and @MangelMaxime

PR #195, opened 2018-12-03
Webpack default minifier

theimowski:
@alfonsogarciacaro is this how you do this?

Issue #194, opened 2018-11-26
Use Webpack default minifier


theimowski:
See https://github.com/SAFE-Stack/SAFE-template/issues/190#issuecomment-441476975

alfonsogarciacaro:
Please see this too: https://github.com/SAFE-Stack/SAFE-template/issues/190#issuecomment-444041124

Also, every time I learn another mystery from Webpack/Babel configuration (and there's something new every day), I'm updating the [webpack-config-template](https://github.com/fable-compiler/webpack-config-template/blob/master/webpack.config.js). I'd recommend to align that template with SAFE-template as much as possible (not sure if using Paket Github dependencies or similar makes sense here). But please note we cannot merge them (as some people may use the webpack-config-template independently) and I can't send a PR here every time I make a change.

theimowski:
How does webpack-config-template relate to versions of the JS dependencies (webpack, loaders, etc)? 

/cc @isaacabraham - it might be good idea to go in pair with this webpack template. To do this however, we'd need to identify and address all differences, as there are some.

isaacabraham:
@theimowski I believe the Fable template is being deprecated in favour of the safe one anyway. Alfonso can you confirm this?

theimowski:
I think these are two separate things:
* [fable template](https://github.com/fable-compiler/fable-templates) is a dotnet template and indeed seems to be depracated now
* [webpack-config-template](https://github.com/fable-compiler/webpack-config-template) is a different GitHub repository which I believe Fable docs point to as the go-to as of now

I'm for the idea of having a single source of truth for weback config, that's why I created #204 - so that future changes to webpack-config-template can be pulled to safe template simply by doing `paket update`

theimowski:
Closed by #204

Issue #152, opened 2018-08-31
"last 2 versions" considered harmfull


0x53A:
see https://jamie.build/last-2-versions

quoting from there:

--------------

## What to do instead

This would be my personal recommendation: 

```js
"browsers": [
  ">0.25%",
  "not ie 11",
  "not op_mini all"
]
```

noppa:
I think lots of people still want to support IE 11.

0x53A:
I think the main message is that ``last two versions`` is broken, regardless of your needs.

If you need IE11, then you could either use ``>0.25%`` which implicitly includes IE11, or better, ``>0.25%, IE11``, which explicitly includes it.

theimowski:
Closed by #204

PR #204, opened 2018-12-19
Unify webpack configuration with webpack-config-template

theimowski:
Idea is to take webpack configuration from https://github.com/fable-compiler/webpack-config-template , adjust a couple of configuration values and include in the template, instead of maintaining a separate different webpack conf


theimowski:
/cc @alfonsogarciacaro @isaacabraham @MangelMaxime want to review?

Related: #195 #194 #152

Content/src/Client/webpack.config.js
```diff
@@ -1,51 +1,80 @@
...
-        "app": [
-            "whatwg-fetch",
-            "@babel/polyfill",
-            resolve("./Client.fsproj")
-        ]
-    },
+    // In development, bundle styles together with the code so they can also
+    // trigger hot reloads. In production, put them in a separate CSS file.
+    entry: isProduction ? {
+        app: [resolve(CONFIG.fsharpEntry), resolve(CONFIG.cssEntry)]
```
MangelMaxime: You can simplify the entries by removing `resolve`.

Webpack expect an absolute path only for the `output.path` node.

Content/src/Client/webpack.config.js
```diff
@@ -58,62 +87,79 @@ module.exports = {
...
+            {
+                test: /\.(png|jpg|jpeg|gif|svg|woff|woff2|ttf|eot)(\?.*)?$/,
+                use: ["file-loader"]
             }
         ]
     }
 };
+
+function resolve(filePath) {
+    return path.isAbsolute(filePath) ? filePath : path.join(__dirname, filePath);
```
MangelMaxime: You can simplify by using:

```fs
function resolve(filePath) {
    return path.join(__dirname, filePath)
}
```

build.fsx
```diff
@@ -32,6 +32,38 @@ Target.create "Clean" (fun _ ->
     Git.CommandHelper.directRunGitCommandAndFail "./Content" "clean -fxd"
 )
 
+Target.create "BuildWebPackConfig" (fun _ ->
+    let srcDir = "paket-files/fable-compiler/webpack-config-template/webpack.config.js"
+    let destDir = "Content/src/Client/webpack.config.js"
+    Shell.copyFile destDir srcDir
+
+    let devServerProxy =
```
MangelMaxime: Is this for supporting project which include a server ?

If we want to lower the complexity, I would expect to use the template feature to choose between 2 `webpack.config.js` file when installing the template.

More or less like it's done for `paket.lock` file

Content/src/Client/webpack.config.js
theimowski: The `webpack.config.js` is now download from webpack-config-template repository. Actually I might consider gitignoring this file to prevent confusion

Content/src/Client/webpack.config.js
theimowski: Same as above - the file comes from webpack-config-template

Content/src/Client/webpack.config.js
MangelMaxime: Ah ok, so this why the `build.fsx` is modifying it to add `api` things.

theimowski: Exactly, as @alfonsogarciacaro [suggested](https://github.com/SAFE-Stack/SAFE-template/issues/194#issuecomment-444042621), we should stick to single source of truth, which I'm happy to do

MangelMaxime: Yes i have no problem with that :)

I just review this PR as asked. We can wait to see if Alfonso wants to modify the original file with my comments or no I guess.

Issue #208, opened 2018-12-30
Deployment for GCP Kubernetes engine


mastoj:
GCP has a super strong kubernetes story, Google Kubernetes Engine (GKE), and it would be nice to be able to deploy there in a simple manner.

The workflow I'm thinking is:

1. Build docker and tag it with the SHA from the last git commit
2. Push the docker image to google cloud container registry
3. Authenticate with google cloud kubernetes cluster
4. Deploy to the kubernetes cluster and expose port using command line options

That is the simplest possible thing to get it up I think. There are probably ways some of it can be generalized so the kubernetes part is similar for different kubernetes hosting providers, but that should be done when support for the second one is added.

Also, the deploy process for GKE is super basic as described above, but functional. I see this as a way for a developer to get started and then tweak to the specific use case.

isaacabraham:
@mastoj this is fantastic - thank you very much! Adding a new deployment option for another cloud platform will be an excellent addition. There's probably a bit of testing needed for this but I think it might well be possible to get this in for v1.

There's also the [sibling repository](https://github.com/SAFE-Stack/docs) which contains the docs for this - something that would need doing for this would be to also document the new option in the template and how it impacts the FAKE script etc.. I'm happy to help you out with some pointers if needed - just let me know.

PR #209, opened 2018-12-30
Add GCP GKE support

mastoj:
Fixes: https://github.com/SAFE-Stack/SAFE-template/issues/208
Docs: https://github.com/SAFE-Stack/docs/pull/91

Adds support for kubernetes in general and Google Cloud Kubernetes Engine in particular. 

- [x] Create docker tag from git sha
- [x] Publish docker image to google container registry
- [x] Auth with kubernetes cluster
- [x] Deploy as new app
- [x] Update existing app

The PR assumes two tools are pre-installed:

* `gcloud` - cli tool to interact with gcp
* `kubectl` - cli tool to interact with kubernetes

Given the above is installed and that you have a gcp project and gke cluster running. You can test this out by running:

```
dotnet new SAFE --deploy gcp-kubernetes
fake run build.fsx -t Deploy
```

Again, this serves as a way to getting started and will suffice for smaller apps.

### DIFF
--- a/Content/build.fsx
+++ b/Content/build.fsx
@@ -210,7 +210,7 @@ Target.create "Publish" (fun _ ->
 )
 
 Target.create "ClusterAuth" (fun _ ->
-    let clusterName = "safe-cluster"
+    let clusterName = Environment.environVarOrDefault "SAFE_CLUSTER" "safe-cluster"
     let authArgs = sprintf "container clusters get-credentials %s" clusterName
     runTool "gcloud" authArgs "."
 )
--- a/Content/build.fsx
+++ b/Content/build.fsx
@@ -57,17 +57,6 @@ let runTool cmd args workingDir =
     |> Proc.run
     |> ignore
 
-let runToolWithOutput cmd args workingDir =
-    let arguments = args |> String.split ' ' |> Arguments.OfArgs
-    let result =
-        Command.RawCommand (cmd, arguments)
-        |> CreateProcess.fromCommand
-        |> CreateProcess.withWorkingDirectory workingDir
-        |> CreateProcess.ensureExitCode
-        |> CreateProcess.redirectOutput
-        |> Proc.run
-    result.Result.Output |> (fun s -> s.TrimEnd())
-
 let runDotNet cmd workingDir =
     let result =
         DotNet.exec (DotNet.Options.withWorkingDirectory workingDir) cmd ""
@@ -82,6 +71,17 @@ let openBrowser url =
     |> ignore
 
 //#if (deploy == "gcp-kubernetes")
+let runToolWithOutput cmd args workingDir =
+    let arguments = args |> String.split ' ' |> Arguments.OfArgs
+    let result =
+        Command.RawCommand (cmd, arguments)
+        |> CreateProcess.fromCommand
+        |> CreateProcess.withWorkingDirectory workingDir
+        |> CreateProcess.ensureExitCode
+        |> CreateProcess.redirectOutput
+        |> Proc.run
+    result.Result.Output |> (fun s -> s.TrimEnd())
+
 let getGcloudProject() =
     runToolWithOutput "gcloud" "config get-value project -q" "."
 
--- a/Content/build.fsx
+++ b/Content/build.fsx
@@ -19,6 +19,9 @@ open Cit.Helpers.Arm
 open Cit.Helpers.Arm.Parameters
 open Microsoft.Azure.Management.ResourceManager.Fluent.Core
 //#endif
+//#if (deploy == "gcp-kubernetes")
+open Fake.Tools.Git
+//#endif
 
 let serverPath = Path.getFullName "./src/Server"
 let clientPath = Path.getFullName "./src/Client"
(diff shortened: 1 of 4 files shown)


mastoj:
This has only been tested on OSX. No Windows machine available at the moment. I think it could work on Windows as well since I'm not using any shell commands.

Content/paket.dependencies
```diff
@@ -64,4 +64,7 @@ group Build
     nuget Fake.IO.Zip
     github CompositionalIT/fshelpers src/FsHelpers/ArmHelper/ArmHelper.fs
 //#endif
+//#if (deploy == "gcp-kubernetes")
+    nuget Fake.Tools.Git
```
theimowski: we'll need to add new case to paket lock generation for this - I can add that afterwards

mastoj: sounds good, since I have no clue about how that part works.

Content/build.fsx
```diff
@@ -137,17 +185,46 @@ Target.create "Bundle" (fun _ ->
...
+    let pushArgs = sprintf "push %s" dockerTag
+    runTool "docker" pushArgs "."
+)
 
-    let tagArgs = sprintf "tag %s %s" dockerFullName dockerFullName
-    runTool "docker" tagArgs "."
+Target.create "ClusterAuth" (fun _ ->
+    let clusterName = Environment.environVarOrDefault "SAFE_CLUSTER" "safe-cluster"
+    let authArgs = sprintf "container clusters get-credentials %s" clusterName
+    runTool "gcloud" authArgs "."
```
theimowski: `ERROR: (gcloud.container.clusters.get-credentials) One of [--zone, --region] must be supplied: Please specify location.`

Content/build.fsx
```diff
@@ -137,17 +185,46 @@ Target.create "Bundle" (fun _ ->
...
+//#if (deploy == "gcp-kubernetes")
+Target.create "Docker" (fun _ ->
+    let dockerTag = createDockerTag dockerImageName
+    buildDocker dockerTag
+)
+
+Target.create "Publish" (fun _ ->
+    let dockerTag = createDockerTag dockerImageName
+    let pushArgs = sprintf "push %s" dockerTag
+    runTool "docker" pushArgs "."
```
theimowski: `unauthorized: You don't have the needed permissions to perform this operation, and you may have invalid credentials. To authenticate your request, follow the steps in: https://cloud.google.com/container-registry/docs/advanced-authentication` <- should be probably reflected in docs?

Content/build.fsx
mastoj: I've updated the docs to say how you set a default zone, with that set it should work. Is that enough?

theimowski: yep, fair enough!

mastoj:
It shouldn't be too hard to add kubernetes support for azure and aws after this I guess. Not sure how that works on those platforms, but I imagine it is similar.

theimowski:
Can you rebase on master?

mastoj:
Yes, that was my plan all along :).

Content/build.fsx
```diff
@@ -19,6 +19,10 @@ open Cit.Helpers.Arm
 open Cit.Helpers.Arm.Parameters
 open Microsoft.Azure.Management.ResourceManager.Fluent.Core
 //#endif
+//#if (deploy == "gcp-kubernetes")
+open System.Text.RegularExpressions
+open Fake.Tools.Git
```
theimowski: Just thinking - is it really crucial to employ Git here? Can we find a different way to create a docker tag? Adding this dependency has 2 main drawbacks:
* it implies the user needs to actually create git repo and commit before testing out the deployment
* makes template a bit harder to maintain

Content/build.fsx
```diff
@@ -52,6 +56,17 @@ let runTool cmd args workingDir =
     |> Proc.run
     |> ignore
 
+let runToolWithOutput cmd args workingDir =
```
theimowski: let's add this only for `gcp-kubernetes`

Content/build.fsx
mastoj: It can be solved with a version variable. The reason I added it is because I see it as a good example of how to version your docker container based on your git sha, which does make some things around versioning slightly simpler in some scenarios, and also deterministic.

If it adds a lot of complexity I can, of course, figure something else out. What do you think based on motivation to why I added it in the first place @theimowski?

mastoj:
Rebased on master now.

Content/build.fsx
theimowski: Yes I totally agree it's a go-to solution for deterministic tagging.
However in case of a template it's more steps for a user to try out functionality, and we strive to make the template as easy to use as possible and with minimal steps to deploy.

Can you push to Google Container Registry a docker image without tag (`projectId/projectName` instead of `projectId/projectName:tag`)? 
If so maybe this could be the easiest way to start with. Otherwise - what would a "version variable" look like?

mastoj:
Yes, that is possible. But that might make the user confused when the try
to make changes and nothing happen since the tag is still the same. The
version could just be something simple as v1 used as rag for the container
image. I can make it as simple as possible and instead write something
about the more advanced tagging in the docs.

theimowski:
> I can make it as simple as possible and instead write something
about the more advanced tagging in the docs.

Please do - that would be my preference.

>  nothing happen since the tag is still the same

Is it really the case? I mean if you push the image with the same tag, doesn't it override the existing one in registry? I think that is what happens in case of Docker Hub for example.

mastoj:
It overrides in the registry, yes, but I don’t think kubernetes consider
the image as new if it has the same tag.

theimowski:
If it doesn't, let's just mention that in docs. For the template what really matters is the first-time deploy.

theimowski:
Thanks for that. I'm aware we might not be contributing to "best practices" here, however IMO it's crucial to find a right balance between simplicity and correctness for this template.

PR #212, opened 2019-01-01
[WIP]: Start work on integration of Azure Storage TP.

### LINKED ISSUE #196: Add Storage Type Provider to Azure template
We should include the Azure Storage TP to the template once it reaches v2 with support for net core.

isaacabraham:
I've got a basic version of the template that now uses Azure Storage for the counter state. It's very simple, but does work.

The only caveat is that you (obviously) need some Azure Storage account to get it to work locally; the obvious choice is the storage emulator. The official one is Windows only but there's an unofficial port which should also work (https://github.com/azure/azurite) although I haven't tested it. Since there's no schema to work with to start from, I've used the schema file approach but included a comment that explains that you can ignore this and use a real storage account for schema instead.

Fixes #196


Content/src/Server/ServerGiraffe.fs
```diff
@@ -1,34 +1,55 @@
...
 let publicPath = Path.GetFullPath "../Client/public"
-//#endif
+#endif
 let port = 8085us
 
-let getInitCounter () : Task<Counter> = task { return { Value = 42 } }
+#if (deploy == "azure")
+// Initialise the Azure storage account with a container and the counter state.
+do
+    safeData.AsCloudBlobContainer(runtimeAzure).CreateIfNotExistsAsync().Wait()
```
bruinbrown: Should we be encouraging people to create containers in code still or should we instead start creating containers within the ARM template? https://samcogan.com/create-azure-storage-containers-with-arm-templates/

isaacabraham: That's a nice new addition to ARM. However, how will this work from a local development perspective (think about the out-of-the-box experience)

theimowski:
What's the status on this?

isaacabraham:
I'll spend time on it this week. Either it'll be done before end of next week or we'll just back it out and push back to post-v1.

isaacabraham:
@theimowski I've simplified the integration point for this so it doesn't actually affect the main counter sample but has the nuget package and storage account ready to go as needed. I've *locally* tested all three implementations and they ran. However, I still want to test the azure deployment itself as well.

Content/src/Server/ServerGiraffe.fs
```diff
@@ -1,35 +1,44 @@
-open System
-open System.IO
-open System.Threading.Tasks
-
+#if (remoting)
+open Fable.Remoting.Server
```
theimowski: I prefer sorting not alphabetically, but rather by modules first, e.g.:

1. .net 
1. asp
1. 3rd party libs
1. own modules

Content/src/Server/azure-schema.json
```diff
@@ -0,0 +1,7 @@
+{
+    "safedata": {
+        "counter.txt": {
```
theimowski: what is counter.txt?

Content/src/Server/ServerGiraffe.fs
```diff
@@ -1,35 +1,44 @@
...
+open Fable.Remoting.Giraffe
+#endif
+#if (deploy == "azure")
+open FSharp.Azure.StorageTypeProvider
+#endif
+open FSharp.Control.Tasks.V2
+open Giraffe
+#if (deploy == "azure")
+open global.Shared
```
theimowski: what is global.Shared ?

Content/src/Server/ServerGiraffe.fs
isaacabraham: Mmmmm. This is subjective, so I'll undo the changes I made here - we can discuss this in another PR  :-)

Content/src/Server/ServerGiraffe.fs
isaacabraham: I seem to recall that I had to do this because there was another `Shared` namespace that was clashing with the Shared namespace in the common types file in the template. `global.Shared` just relates to that. I'll check what caused it.

isaacabraham:
@theimowski I've not had the opportunity to get this done in time this week. Therefore I think we should remove this from the v1 milestone and put it in afterwards.

PR #218, opened 2019-01-18
Update to netcore 2.2

mvsmal:
This PR updates SAFE template to .NET Core 2.2.

I also had to update lock files, which caused the update to FAKE 5.12 and other packages updates.


mvsmal:
I had to update Travis environment to Xenial, because Trusty's dpkg is outdated.

mvsmal:
@theimowski In my projects I usually ignore it, since it is an autogenerated by Ionide binary file for background symbol caching.

@Krzysztof-Cieslak, should it be committed to git, actually?

theimowski:
Thanks @mvsmal !

PR #226, opened 2019-02-01
Add IIS deployment option

Zaid-Ajaj:
> I wanted to open an issue first but PR's are nice when they come out of nowhere ;) 

This PR adds `--deploy iis` to the template. There are many problems when it comes to hosting IIS but once you get them all right, it is really nice to work with! 

### Problem 1: Virtual Paths
Running a web app inside IIS means it will probably run inside a virtual path i.e. http://localhost/SafeApp. This behavior breaks routing because if you `fetch "/api/counter"` the request goes to `http://localhost/api/counter` which is the root IIS application. The fix is to normalize all requests made to the server by prepending the virtual path to the route, so I added a little module on the client to do so:
```fs
module ServerPath = 
    open System
    open Fable.Core
    
    /// when publishing to IIS, your application most likely runs inside a virtual path (i.e. localhost/SafeApp)
    /// every request made to the server will have to account for this virtual path
    /// so we get the virtual path from the location
    /// `virtualPath` of `http://localhost/SafeApp` -> `/SafeApp/` 
    [<Emit("window.location.pathname")>]
    let virtualPath : string = jsNative 

    /// takes path segments and combines them into a valid path
    let combine (paths: string list) =
        paths 
        |> List.map (fun path -> List.ofArray (path.Split('/')))
        |> List.concat 
        |> List.filter (String.IsNullOrWhiteSpace >> not) 
        |> String.concat "/"
        |> sprintf "/%s"

    /// Normalized the path taking into account the virtual path of the server
    let normalize (path: string) = combine [virtualPath; path]
``` 
so now you have to call `ServerPath.normalize` for every HTTP request:
```fs
fetch (ServerPath.normalize "/api/counter") // etc.
```

I will need to add this to the docs
### Problem 2: Hosting Suave:
No idea how to implement this yet because this deployment option relies on `AspNetCoreModule` but Suave runs standalone so I am not sure how to implement it (running Suave on Kestrel maybe?) for now only Saturn and Giraffe options are implemented

### Problem 3: netcore2.2:
Netcore2.2 requires "Out of Process" model and `AspNetCoreModule` instead of the new and highly performant `AspNetCoreV2` and the "In Process" model, this is a known issue:
```xml
    <PropertyGroup>
        <OutputType>Exe</OutputType>
        <TargetFramework>netcoreapp2.2</TargetFramework>
        <!-- cannot use InProcess with netcoreapp2.2  -->
        <!-- see https://github.com/dotnet/core/blob/master/release-notes/2.2/2.2-known-issues.md -->
        <AspNetCoreHostingModel>OutOfProcess</AspNetCoreHostingModel>
        <AspNetCoreModuleName>AspNetCoreModule</AspNetCoreModuleName>
    </PropertyGroup>
```

This is PR is ready to review and I have testing multiple combinations and they all work out (on my local machine) 

The documentation on how to setup IIS and actual deployment is coming up :)


isaacabraham:
You can run Suave on IIS because it works on Azure's App Service which is effectively IIS as a service. YOu need to override the HTTP module in IIS to redirect all traffic over to the web server - see the web.config in the SAFE template when you create it with Suave and Azure configuration.

Zaid-Ajaj:
I am having been trying to get Suave to run properly from IIS with no luck, first of all `HttpPlatformHandler` cannot find the compiled `Server.dll`  because of the relative path set in web.config (i.e. `./Server.dll`) after an hour of investigation it seems that the "current directory" of the application is set to `C:\Windows\System32\inetsrv`. this is where w3wp.exe lives and from which the `dotnet` child process is initiated to get the server running. So if I change web.config to use the absolute path of where `Server.dll` is compiled I get the sever running but I still have problems because it cannot find the static content: dotnet thinks they are relative to the parent process (IIS) and not the child process (dotnet) so Suave looks for static content inside `C:\Windows\System32\inetsrv\Client\path`

Aside from that, the API route `/api/init` is not being hit for some reason, I only get the HttpNotFound handler from Suave that returns "Not Found!" 

>  because it works on Azure's App Service which is effectively IIS as a service

I am pretty sure Azure does fancy stuff to make everything "just work" because azure can assume a single application per app service so it can simply copy everything to where IIS is installed

isaacabraham:
Hmm. Not sure about that last part - the app just lives in wwwroot. Don't think there's any copying going on at all.

Zaid-Ajaj:
@isaacabraham Maybe not copying things around but after a long day of testing different stuff, I confirmed that that the only way to get Suave running under IIS is to have it as the default website which is why it "just works" on Azure. 

Suave app must b running as root level application in IIS because otherwise the routing will never work using the proxy, here is why:
Suave has routes filters defined as follows:
```
GET >=> path "/" >=> OK "index" 
```
When running this inside a virtual application which is most likely the case for manually managed IIS servers, for example
```
http://localhost/SafeApp
``` 
then IIS will proxy the requests made with the prefix `/SafeApp` and the routes will never hit because they don't expect that prefix in the path. This is why the app must be running on root or the application has to have information internally about the virtual path (yeah we don't want that either).

I will add the required Suave-specific code and explain in the docs that Suave must be running as a root website inside IIS

Zaid-Ajaj:
This PR is ready to merge, detailed docs for the deployment are pushed in [another PR](https://github.com/SAFE-Stack/docs/pull/100) to SAFE docs and are ready to merge as well @theimowski @isaacabraham @Krzysztof-Cieslak

theimowski:
Thanks @Zaid-Ajaj I'll review this soon

theimowski:
We're currently narrowing down to the v1 release for the template.
This IIS integration is surely nice and thanks a lot @Zaid-Ajaj for hard work put into that. 
We want it in, however I think we should put it on hold and wait until we release SAFE template 1.0 

After 1.0 is released we'll consider placing certain functionalities into separate packages and IIS seems a good fit for that.
How does that work for you @Zaid-Ajaj ?

Zaid-Ajaj:
Thanks a lot for the review @theimowski, I am glad to hear that this will PR be merged, I am not in a hurry and will happily wait until the template is stable before merging :smile:

Zaid-Ajaj:
Congrats on the stable release of SAFE 🎉🎉 

I was wondering if you are still interested in supporting IIS and this PR? it would help me (and I guess a large group of other people) to introduce SAFE to their organisation where IIS is still the preferred way of deploying apps

theimowski:
Thanks @Zaid-Ajaj 
Yes surely we'd like to integrate this - can you work on rebasing that on recent master?
Also would be great if you could provide steps for end to end testing of this option for someone not running on Windows.

Zaid-Ajaj:
> can you work on rebasing that on recent master?

Done! (hopefully I did the rebase correctly, figers crossed...)

> Also would be great if you could provide steps for end to end testing of this option for someone not running on Windows.

That would be a bit difficult as non-Windows based IIS (such as XSP4 with mono) can behave differently than the "real" IIS on Windows. To properly test this feature you would need Windows I am afraid

theimowski:
Looks like CI failed on building project with `--deploy iis` option - can you check that?

> To properly test this feature you would need Windows I am afraid

Right, do you think it would be reasonable to add a Vagrant-based environment (preferably on virtualbox) with IIS provisioned?

Zaid-Ajaj:
It is failing with `--deploy iis` and `--server suave`, I removed the `Bundle` target that seems to be included twice, maybe that is the cause? In any case during dev. the `run` target should work properly without actually deploying anything to iis 

> Right, do you think it would be reasonable to add a Vagrant-based environment (preferably on virtualbox) with IIS provisioned?

TBH I haven't worked with Vagrant before so I can't say for sure.

Zaid-Ajaj:
OK something is not right, using:
```
dotnet new SAFE --server suave --deploy gcp-kubernetes --layout fulma-login --communication remoting
```
gives this error:
```
Unhandled Exception: System.IO.FileLoadException: Could not load file or assembly 'Suave, Version=2.5.0.0, Culture=neutral, PublicKeyToken=null'. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
```
I seen this before on my local machine in a non-SAFE project when using remoting and suave, I still don't know why this happens because `Suave 2.5.0` *should* be there as it is a dependency of `Fable.Remoting.Suave` 

maybe @forki can shed some light on this matter?

Zaid-Ajaj:
OK build is successful, but still there was that exception here above, @theimowski can you please take a look at the build logs?

theimowski:
Not sure why the diff shows changes already present in master.. The FileLoadException for Suave seems to be unrelated - I can see it in master builds as well

theimowski:
let's get the ball rolling - released in 1.2.0 thanks @Zaid-Ajaj

Zaid-Ajaj:
@theimowski Thanks a lot, really appreciate it 🙏 merging the [docs PR](https://github.com/SAFE-Stack/docs/pull/100) would be nice too

PR #235, opened 2019-03-08
Remove ionide folder from git.

isaacabraham:
The ionide cache file shouldn't be committed, it'll just grow and slow down pulls, plus it causes pain resetting because Ionide holds a lock on the file etc.


theimowski:
Agreed - thanks. I think it was added by mistake back in a day

PR #241, opened 2019-03-29
Fix initial build failure

isaacabraham:
Using paket global, a `dotnet new safe` fails, I think because paket restore targets are out of date. I've updated them using the latest one that is generated by paket and this *seems* to fix it.

Can someone else test locally though please :-)


isaacabraham:
Fixes #106 I think?

theimowski:
Yup - thanks that solves the issue!

PR #243, opened 2019-04-06
Fix gcloud tool name on Windows

sWW26:
Fixes an issue where the gcloud  command is not found on Windows.

Followed the same pattern that is applied to other tools.

### DIFF
--- a/Content/build.fsx
+++ b/Content/build.fsx
@@ -46,6 +46,9 @@ let npxTool = platformTool "npx" "npx.cmd"
 //#else
 let yarnTool = platformTool "yarn" "yarn.cmd"
 //#endif
+//#if (deploy == "gcp-kubernetes")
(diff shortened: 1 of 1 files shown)


Content/build.fsx
```diff
@@ -46,6 +46,9 @@ let npxTool = platformTool "npx" "npx.cmd"
 //#else
 let yarnTool = platformTool "yarn" "yarn.cmd"
 //#endif
+//#if (deploy == "gcp-kubernetes")
```
theimowski: need to add ` || deploy == "gcp-appengine"` here

theimowski:
`gcloudTool` needs to be present for `gcp-appengine` option as well - see my inline comment

theimowski:
Thanks!

Issue #247, opened 2019-04-14
Migrate from SASS to SCSS


hravnx:
If I add a style to the (initially empty) style.sass file, it will fail to compile.

To reproduce:

1. Create an empty folder and navigate into it
2. Run `dotnet new SAFE -lang F#`
3. Run `fake build --target run` and wait for the browser to come up and present the counter example.
4. Open the `src/Client/style.sass` file in the editor of your choice and add a style to it. For example:
```sass
h3 {
  width: 100px;
}
```
5. Save it. Observe the error in the compilation process:
```
 ERROR in ./src/Client/style.sass (./node_modules/css-loader/dist/cjs.js!./node_modules/sass-loader/lib/loader.js??ref--6-2!./src/Client/style.sass)
Module build failed (from ./node_modules/sass-loader/lib/loader.js):

h3 {
  ^
     Expected newline.
h3 {
   ^
      stdin 1:4  root stylesheet
      in /Users/username/Projects/Spikes/sss/src/Client/style.sass (line 1, column 4)
 @ ./src/Client/style.sass 2:14-136 21:1-42:3 22:19-141
 @ multi (webpack)-dev-server/client?http://localhost:8080 (webpack)/hot/dev-server.js ./src/Client/style.sass
``` 

OS: macOS 10.14.4
SAFE template: 1.0.1
dotnet core:  2.2.202

Zaid-Ajaj:
Hmm no sass expert here but the [docs](https://sass-lang.com/guide) say the syntax is a bit different:
```sass
h3
  width: 100px
```

hravnx:
You are right, of course - I normally use scss syntax. Silly of me, thanks :-)

Zaid-Ajaj:
No problem! I think the template also includes a CSS loader, so you can create a plain old `.css` file and import it like this in your entry file on the `Client` project:
```fs
import "./custom-styles.css" 
```
of course assuming `custom-styles.css` is a sibling file to entry file that contains the `import` statement

hravnx:
I just renamed `style.sass`to `style.scss` and changed the `cssEntry` property in `webpack.config.js` and happiness ensued. Thanks again for the quick help!

theimowski:
I choose the SASS over SCSS as I saw it was being used somewhere in the sample Fable projects. @alfonsogarciacaro @MangelMaxime do you have an opinion on whether any is preferred nowadays?

alfonsogarciacaro:
[...]

PR #248, opened 2019-04-16
Change default from sass to scss

hravnx:
As per issue [247](https://github.com/SAFE-Stack/SAFE-template/issues/247) it was decided to change the default style syntax from the old sass to the current scss.

This PR does that by changing the name and content of the old `src/Client/style.sass` file and by changing the corresponding cssEntry property in the generated webpack config file.

The sass content was changed to scss by the sass-convert utility built into the official sass gem.


theimowski:
Great, thank you! Let the CI run tests and then I'll make a quick verification tests for some of the fulma templates.

theimowski:
released in 1.1.0

PR #250, opened 2019-04-24
New fable dependencies

theimowski:


Zaid-Ajaj:
Update remoting dependencies as well? Latest updates include binary transfer optimizations (no public api change) 
 
Fable.Remoting.Client ~> 5 (5.3.0)
Fable.Remoting.Giraffe (3.7.0)
Fable.Remoting.Suave (3.8.0)

theimowski:
Yeah makes sense, server dependencies also need to be updated (Giraffe and Suave)?

Zaid-Ajaj:
> server dependencies also need to be updated

Yes please, always use latest client with latest server (although latest server is compatible with older client versions)

Content/package.json
```diff
@@ -10,8 +10,8 @@
         "copy-webpack-plugin": "^4.6.0",
```
forki: can we get rid of explicit reference of core-js? @alfonsogarciacaro

Content/paket.dependencies
```diff
@@ -18,9 +18,9 @@ group Server
 //#elseif (!remoting && server == "suave")
     nuget Thoth.Json.Net
 //#elseif (remoting && server == "suave")
-    nuget Fable.Remoting.Suave
+    nuget Fable.Remoting.Suave ~> 3
```
forki: urgs explicit version ranges!?

Content/src/Client/Client.fs
```diff
@@ -70,34 +67,48 @@ module Server =
...
       Remoting.createApi()
       |> Remoting.withRouteBuilder Route.builder
       |> Remoting.buildProxy<ICounterApi>
-    #endif  
-
-
+    #endif
+#else
+// Fetch a data structure from specified url and using the decoder
+let fetchWithDecoder<'T> (url: string) (decoder: Decoder<'T>) (init: RequestProperties list) =
```
forki: wonder if this belongs in some lib!?

Content/package.json
theimowski: https://github.com/fable-compiler/webpack-config-template/issues/12#issuecomment-490862309

forki:
Weird. I was able to remove it

Tomasz Heimowski <notifications@github.com> schrieb am Di., 14. Mai 2019,
11:48:

Content/paket.dependencies
theimowski: That makes it easier to update lock files without getting breaking changes (as long as the deps follow semantic versioning)

Content/src/Client/Client.fs
theimowski: @alfonsogarciacaro @MangelMaxime I took it from fable migration guide

MangelMaxime: @forki Well, personally I rewrote my own version in my application because I need to handle refresh tokens etc.

In a previous issue https://github.com/MangelMaxime/Thoth/issues/130, I mentioned I could create a `Thoth.Fetch` package to host this kind of helpers. If you think it is needed, I can do it.

forki:
It just looks like something that's better for in a lib than inside of app
code

Maxime Mangel <notifications@github.com> schrieb am Di., 14. Mai 2019,
12:46:

MangelMaxime:
@forki No problem, 
I should do it today or tomorrow.

MangelMaxime:
@forki @theimowski 

I am designing the API for the Fetch library and here is my proposition:

```fs
open Thoth.Fetch

// The decoder is provided by the user (manual or generated before)
promise {
    let! res = Fetch.fetchAs("http://localhost:3000/posts/1", Post.Decoder)
    // ...
}

// The decoder retrieved from the cache and if not found, it will be generated and cached
promise {
    let! res = Fetch.fetchAs<Post>("http://localhost:3000/posts/1", isCamelCase = true)
    // ...
}
```

The `isCamelCase = true` argument is here to show that I am allowing passing arguments to `Thoth.Json` in this case to control the representation of the keys in the objects.

The `(init: RequestProperties list)` is also optional and can be provided by the user if needed.

Content/src/Client/Client.fs
theimowski: I'm happy to consume `fetchAs` function from a library - indeed that seems like unnecessary boilerplate in application code

Content/package.json
alfonsogarciacaro: Yes, unfortunately it seems that now you need to make the core-js version explicit **both** in package.json and webpack.config.json :/

Content/src/Client/Client.fs
alfonsogarciacaro: I think it's better to keep Fable.Fetch and Thoth.Json packages independent and don't create a hard dependency between them just to enable a single function.

alfonsogarciacaro: The solution could be to create an extra package that has the dependency on both Fable.Fetch and Thoth.Json. Although I'm still in doubt if we should insist on Fable.Fetch or try something more idiomatic like Fable.SimpleHttp (which is also compatible with IE 11 without a polyfill).

forki: @alfonsogarciacaro I think @MangelMaxime proposed a third package which marries both

alfonsogarciacaro: Ups, I just noticed you were continuing the discussion below. If Maxime is already on it, that's fine :)

forki: so in other words: both packages stay independent. but if you want to use both anyway then you use the married thing and it will introduce the original packages as transitives

MangelMaxime: Yes, I am working on Thoth.Fetch package which offers an API over Fable.Fetch + Thoth.Json.

Like that we still keep the independent low-level packages.

MangelMaxime:
@forki @theimowski 

I just published **Thoth.Fetch** version `1.0.0-beta-001`. Can you please give it a try? 

If you confirm that the package is working, I will release it as stable. I will also finish the documentation of the code before stable version. I just wanted to provide a beta version as quickly as possible.

If you are not sure how to use the library, you can take a look at the tests.

forki:
LGTM. Works already nicely in my app

MangelMaxime:
@forki @theimowski 

Ok, I just released Thoth.Fetch 1.0.0

forki:
yay

theimowski:
Pushed a change using the new lib

theimowski:
Released in v1.6

PR #254, opened 2019-05-03
Add devocntainer definition

Krzysztof-Cieslak:
Having in the devcontainer definition in workspace will make VSCode (if user have `remote-containers` extension installed) prompt user to open the workspace inside the defined docker container.
As you can see in the docker container we have .Net SDK 2.2, Node, Yarn, Paket and FAKE - i.e all our requirements.
VSCode will also automatically install the extensions specified in `devcontainer.json`.

All this means users can start developing applications with VSCode and docker as only requirements - and they will automatically get full configure development environment with all dependencies installed.

### DIFF
--- a/Content/.devcontainer/Dockerfile
+++ b/Content/.devcontainer/Dockerfile
@@ -1,0 +1,28 @@
+FROM mcr.microsoft.com/dotnet/core/sdk:2.2
+
+# Add keys and sources lists
+RUN curl -sL https://deb.nodesource.com/setup_11.x | bash
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
+RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" \
+    | tee /etc/apt/sources.list.d/yarn.list
+
+# Install node, 7zip, yarn, mono, git, process tools
+RUN apt-get update && apt-get install -y nodejs p7zip-full yarn git procps
+
+# Clean up
+RUN apt-get autoremove -y \
+    && apt-get clean -y \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install fake
+RUN dotnet tool install fake-cli -g
+
+# Install fake
+RUN dotnet tool install paket -g
+
+# add dotnet tools to path to pick up fake and paket installation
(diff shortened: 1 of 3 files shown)


Krzysztof-Cieslak:
> 
> 
> ```
> # Install fake should be 	# Install paket
> ```

Thanks, fixed

Content/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM mcr.microsoft.com/dotnet/core/sdk:2.2
+
+# Add keys and sources lists
+RUN curl -sL https://deb.nodesource.com/setup_11.x | bash
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
+RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" \
+    | tee /etc/apt/sources.list.d/yarn.list
+
+# Install node, 7zip, yarn, mono, git, process tools
```
AndreasHassing: You're not installing mono, are you?

ScottHutchinson:
Title "Add devocntainer definition" should be "Add devcontainer definition"

theimowski:
image

That's probably because the `Remote - Containers` extension is in preview version - wondering whether we should wait with that until it's in a stable version? I don't think many users have the insider version of VS Code installed (I didn't)

Krzysztof-Cieslak:
> wondering whether we should wait with that until it's in a stable version?

TBF, I think we should merge this despite being insiders-only. (But I'm super excited about this feature)

If someone has insiders version then it is just added value. 
If someone has stable then those are just 3 small files they can ignore. And, what's important, people will already have this definition in their repos when the feature will got released in stable. I.e. they'll be able to use it without any additional work. 

And yes, we definitly should add extension recommendation.

Nhowka:
I'm behind a corporate proxy and couldn't make it work. The dotnet tools were installed, but at the end it's trying to connect to my host machine name on port 80 and it's failing. Not sure what is trying to connect or what do I have to do...

theimowski:
Ok agreed.

Does the `Debugger for Chrome` extension needs to be installed locally or remotely? I can't seem to get it working using `Debug SAFE app`:

image

Issue #255, opened 2019-05-08
Use a Shared project instead of including shared files individually


theprash:
I propose adding a `Shared/Shared.fsproj` file, referenced from Server and Client projects. Even though there is only one shared file in the initial template, I have seen projects take this template and add many shared files in a way that doesn't scale very well.

With a Shared project file, if there are multiple shared files, they don't have to be included in both Server and Client in the correct order: There is one source of truth for file ordering.

Is there any reason this isn't the approach already taken by the template?

theimowski:
I've been thinking about that back in a day - just wasn't sure if that scenario is supported by Fable.
But if that works I'm all in for that change

theprash:
Can this be re-opened? The change was reverted due to #258, which seems to be an issue only affecting Linux.

Using Shared files like this still feels very wrong to me. We are teaching people that this is how to handle project dependencies in F#. I have just seen someone do this between two .NET assemblies within a SAFE project.

@iNecas, could you help us solve this? Does using a shared project between Fable and .NET still cause issues for you?

theprash:
#258 seems to be caused by this Fable bug: https://github.com/fable-compiler/Fable/issues/1953. It looks like it was happening intermittently and it's mostly stopped happening. So I'm going to open another PR making this change again.

theimowski:
Yes, I'm all in for having Shared as an fsproj - actually that's what has been done in the v2_minimal branch

PR #256, opened 2019-05-08
Add Shared.fsproj

theprash:
See #255.

### DIFF
--- a/Content/src/Server/Server.fsproj
+++ b/Content/src/Server/Server.fsproj
@@ -11,7 +11,7 @@
         #endif
     </PropertyGroup>
     <ItemGroup>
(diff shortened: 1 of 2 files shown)


theimowski:
Thanks!

iNecas:
FYI this seemed to break the template, at least on my Linux machine https://github.com/SAFE-Stack/SAFE-template/issues/258

theimowski:
Had to revert the change
It broke on my Mac as well - the fix was to clear fable caches (.fable directory)
We need to make it work in all scenarios

theimowski:
@theprash  any chances you can have another look at this? would be nice to have it

PR #262, opened 2019-05-16
Set Fable variable on client project

forki:
this makes the ionide aware of the fact that this is a fable project.
This is needed if `#if FABLE_COMPILER` directives that are sometimes needed for SSR - especially when you want to have Thoth.Fetch in one end and Thoth.Net on the other.

/cc @Krzysztof-Cieslak @isaacabraham @MangelMaxime

### DIFF
--- a/Content/src/Client/Client.fsproj
+++ b/Content/src/Client/Client.fsproj
@@ -2,6 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
         <TargetFramework>netstandard2.0</TargetFramework>
+        <DefineConstants>FABLE_COMPILER</DefineConstants>
     </PropertyGroup>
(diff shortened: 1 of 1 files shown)


forki:
For clarification: Fable itselft sets the variable during compilation. So it always worked fine. It's just that there are edge cases where the IDE shows the wrong branch and that's consfusing as fuck

theimowski:
I haven't yet used SSR, but if that helps then I'm happy to pull that in.

Just wondering whether we might have a more general IDE-related issue with those constants:
Say you use them in `Shared.fs` which is shared between Client and Server - should IDE respect constants defined in Client or Server project?

I have a similar issue in commercial project, but based on the OS-specific constants (`#if WINDOWS`, `#if OSX` etc..)

Issue #203, opened 2018-12-18
Clean up after FontAwesome migration


0x53A:
There are still a few unresolved comments in https://github.com/SAFE-Stack/SAFE-template/pull/199.

This is mostly a reminder to myself, but anyone is welcome to fix it.

theimowski:
@0x53A is this still relevant?

theimowski:
ping

PR #264, opened 2019-05-21
Fontawesome small changes

0x53A:
fixes https://github.com/SAFE-Stack/SAFE-template/issues/203

----------

After #199:

image

image

Should actually be:

image

image


0x53A:
Oops, I was apparently braindead when comitting ...

Thank you. I think this is ready.

theimowski:
Ok thanks!

Issue #107, opened 2018-06-21
Add Elmish.Bridge as an option 


theimowski:
Add https://github.com/Nhowka/Elmish.Remoting option as an alternative to Fable.Remoting.

Naming of these two may be misleading, see https://github.com/SAFE-Stack/SAFE-BookStore/pull/342 and https://github.com/Nhowka/Elmish.Remoting/issues/1 for discussions.

The option should be mutually exclusive with `--remoting` - probably would need to change type of this parameter from boolean to choice and rename it accordingly

theimowski:
adding milestone 1.0 as changing parameter's name is a breaking change

Nhowka:
Even with the small template there is still space for both of them to work together. The initial state could still be fetched from the server via Fable.Remoting and the new state could be calculated on the server and returned back to the client via Elmish.Remoting. That wouldn't break the existing API and would reinforce the fact that both can work together.

I'll try to make an example and release the project under a new name on this weekend.

Nhowka:
[Renaming is ready!](https://github.com/Nhowka/Elmish.Bridge)

isaacabraham:
@theimowski we should definitely get this in. However, playing devils advocate here, does this *need* to be in for the 1.0 milestone?

theimowski:
@isaacabraham we can go 1.0 without that, but I'd change the parameter from `--remoting` to something more generic

Nhowka:
@theimowski Now with the new name you could keep `--remoting` for Fable.Remoting and have `--bridge` for Elmish.Bridge. No need for breaking changes.

theimowski:
that would mean support for both options - not sure if that's viable?

Nhowka:
@theimowski That https://github.com/SAFE-Stack/SAFE-template/issues/107#issuecomment-399283763 gives an example of what the template would look like with both options, one is not exactly a replacement for the other.

isaacabraham:
@Nhowka how likely is it that people will use both? It will simply matters greatly in the template to have them mutually exclusive.

Nhowka:
@isaacabraham I have no information about someone using Elmish.Bridge, so it's all about my opinion and design abilities. I'm planning to use both on a project that involves inventory management for a supermarket.

Bridge is stateful, so I'll use it in a way to make it possible for two or more people to load the same cargo into it, with real-time updates on all workstations. Having the state on the server is something good. I'll see who has which invoice open and send updates as the items are unloaded from the truck.

[...]

PR #265, opened 2019-05-22
[WIP] Adding Elmish.Bridge as a communication option

Nhowka:
Implementing #107 

I tried to modify as little as possible, only adding a simple clock that can be paused so the tick won't be sent to the client. Now probably need some more comments so the user can know better where to modify.

Sending the PR as the CI is very nice on checking if I broke anything.

cc @theimowski @isaacabraham @Zaid-Ajaj


theimowski:
First thing - big thanks for that, adding a feature like that is quite a challenge and I really appreciate your work on that since I realise what was involved.

Regarding the changes: can we just send the initial counter via Elmish.Bridge instead of introducing the clock?
I'd prefer we remain minimalistic with regards to the communication.

Also on a different note: those changes are quite big, and TBH I have never used Elmish.Bridge before. 
Are you willing to help if any issues related to Elmish.Bridge arise as well as when we need to update dependencies to newer versions?
We just need to make sure we're able to maintain this option after we deliver it.

Zaid-Ajaj:
All in all it looks really good! One thing I noticed is executing side-effects inside event handlers, i.e. calling `Bridge.Send Msg` inside `OnClick` handlers which makes the `view` function impure/not unit-testable. It is not *that* of a big deal but it is better to put such code inside commands as they are responsible for all the side-effects

Nhowka:
> can we just send the initial counter via Elmish.Bridge instead of introducing the clock?

We could, but it's not really suitable for actions that are just one-time things as it creates a stateful connection between the server and client. Instead I could make the counter server-side and the client would just show the counter that would be incremented/decremented on the server, but that would make it harder to add a new endpoint to the api or be compatible with the remoting option as it would not have a starting point for them. Let me know if you are ok with the drawbacks.

> Are you willing to help if any issues related to Elmish.Bridge arise as well as when we need to update dependencies to newer versions?

Yes, I'm willing to work on it and to smooth any rough edges that may appear.

> it is better to put such code inside commands as they are responsible for all the side-effects

After making excuses for my laziness while replying I thought of a smarter way to fix that without much work. I'll test it and apply the changes if it works.

Zaid-Ajaj:
> After making excuses for my laziness while replying I thought of a smarter way to fix that without much work. I'll test it and apply the changes if it works.

It shouldn't require a lot of work to create a command from the `Send` call, just this
```fs
module Cmd = 
  let bridgeSend (msg 't) : Cmd<'t> = [ fun _ -> Bridge.Send msg |> ignore ]
```
There might an existing `Cmd` helper that runs a function but I am not sure

Nhowka:
I might've just changed one problem for another. My idea was creating a new case on the messages, something like `Outgoing of ServerMsg` and then calling `Bridge.Send` with those messages on `update`. 😅

I'll investigate if the `Cmd` helper works ok as `Bridge.Send` depends on the type of the message being sent. But I really like the idea, that would make it more alike the other libraries.

Zaid-Ajaj:
> I'll investigate if the Cmd helper works ok as Bridge.Send depends on the type of the message being sent. 

Yeah then you need to `inline` the function. 

> But I really like the idea, that would make it more alike the other libraries

Yeah definitely! This is a suggestion for the library (maybe not the best place to discuss it here but) the `Bridge.Send` call doesn't communicate back whether the call was successful or not. Maybe something like an `onError : exn -> 'msg`  callback would be nice:
```fs
module Cmd = 
  let inline bridgeSend (msg: 'Msg) (onError:  exn -> 'Msg) = 
    [ fun dispatch -> Bridge.SendSafe msg (onError >> dispatch) ]
```
where `Bridge.SendSafe` has the type: `'Msg -> (exn -> 'Msg) -> unit`

The example above only communicates the errors back to the dispatch loop but you could have other callbacks that tell the success status etc. Let me know what you think

Nhowka:
> Yeah then you need to `inline` the function.

I works! ❤

> `Bridge.Send` call doesn't communicate back whether the call was successful or not

Maybe making a happy one that will gladly ignore the unsent messages when the socket is down and another that will try to add something to the loop. Adding an optional `unit -> unit` to `Bridge.Send` would support that. That would be needed as the dispatch reference inside bridge only points to the top-level message.

```fs
module Cmd =
    let inline bridgeSend (msg:'server) : Cmd<'client> = [ fun _ -> Bridge.Send msg ]
    let inline bridgeSendOr (msg:'server) (fallback:'client) : Cmd<'client> = [ fun dispatch -> Bridge.Send(msg, fun () -> dispatch fallback) ]
```
That is assuming that the message only fails to be sent when the socket closes. That would support different recoveries as different messages are tried to be sent when the socket is not up, but there is already an option to send a message when the socket closes.

theimowski:
@Nhowka  how about sending the clock from server, but use it for `Counter` in the Client model? I.e. you could take the unix epoch seconds and use them as Counter on the client side.
The `Increment` and `Decrement` buttons could work as previously: +/- 1 and then on next clock tick you'd get a value from server which overwrites your clicks.
Or we could hide Increment and Decrement when Elmish.Bridge is used

theimowski:
Can we close this as #267 went in?

Nhowka:
Yes, thanks!

PR #266, opened 2019-05-22
Move .sln to the root directory and add files

### LINKED ISSUE #260: Suggestion: `.sln` file in root directory instead of `src`
I was surprised after creating a new app with `SAFE-template`, to have the solution file in `src` dir instead of the `root` one, as seems to be a standard in the fsharp world (https://github.com/fsharp/fsharp, https://github.com/fsprojects/FSharp.Formatting, https://github.com/fsharp/FAKE, https://github.com/fsprojects/Paket).

This is not only cosmetic thing, but actually makes it harder to work with in real world, as some IDEs (at least Rider), consider the directory with `.sln` file the root dir, with consequences such as FAKE and paket scripts not actually showing up in the IDE.

Any reason for doing it this way and any chance to change it's location?

theprash:
Fixes #260. I added paket files to the solution and project files plus other files that might be useful.

### DIFF
--- a/Content/.template.config/template.json
+++ b/Content/.template.config/template.json
@@ -15,6 +15,7 @@
     },
     "identity": "SAFE.Template",
     "shortName": "SAFE",
+    "sourceName": "SAFE.Template",
     "symbols": {
         "server": {
             "type": "parameter",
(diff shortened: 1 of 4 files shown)


theimowski:
Perfect, thanks!

iNecas:
Awesome work @theprash , thanks also for quick release @theimowski

PR #267, opened 2019-05-27
[WIP] Adding Elmish.Bridge as a communication option (counter version)

Nhowka:
Alternative to #265. Makes the counter server-side with a fallback function so the client can work even when the server disconnects.

Changes made mainly to the elmish part of it.


theimowski:
Yeah I think I prefer this lighter option rather than #265
@Zaid-Ajaj wanna help reviewing?

Content/.template.config/template.json
```diff
@@ -83,13 +83,21 @@
                 {
                     "choice": "remoting",
                     "description": "add Fable.Remoting (https://zaid-ajaj.github.io/Fable.Remoting/) to server and client"
+                },
+                {
+                    "choice": "bridge",
```
theimowski: lets add it to Tests as in previous PR

Content/src/Client/Client.fs
```diff
@@ -88,6 +94,9 @@ module Server =
       |> Remoting.buildProxy<ICounterApi>
     #endif
 let initialCounter = Server.api.initialCounter
+#elseif (bridge)
```
theimowski: this can be removed?

Content/src/Client/Client.fs
```diff
@@ -98,19 +107,22 @@ let initialCounter () = Fetch.fetchAs<Counter> "/api/init"
...
+let init () : Model * Cmd<Msg> =
+    { Counter = None }, Cmd.none
 #else
 // defines the initial state and initial command (= side-effect) of the application
 let init () : Model * Cmd<Msg> =
     let initialModel = { Counter = None }
     let loadCountCmd =
 #endif
-#if (!reaction && remoting)
+#if (!reaction && remoting && !bridge)
```
theimowski: those 3 conditions got bit complex, can we refactor those to make it cleaner? Maybe use nested `#if` statements?

Content/src/Client/Client.fs
```diff
@@ -151,15 +178,29 @@ let update (msg : Msg) (currentModel : Model) : Model =
...
+        #endif
+    #if (bridge)
+    | _, Remote(SyncCounter counter) ->
+        { currentModel with Counter = Some counter}, Cmd.none
+    #else
     | _, InitialCountLoaded initialCount->
         let nextModel = { Counter = Some initialCount }
         nextModel, Cmd.none
-
+    #endif
```
theimowski: don't forget to add your credits to `safeComponents` !

Content/src/Server/ServerGiraffe.fs
```diff
@@ -49,6 +75,8 @@ let webApp =
     |> Remoting.buildHttpHandler
 
 #else
+let getInitCounter () : Task<Counter> = task { return { Value = 42 } }
```
theimowski: let's inline `getInitCounter` both for remoting and the default option - we no longer need it to be a separate function

Content/src/Server/ServerSaturn.fs
```diff
@@ -47,6 +73,8 @@ let webApp =
     |> Remoting.buildHttpHandler
 
 #else
+let getInitCounter() : Task<Counter> = task { return { Value = 42 } }
```
theimowski: as above - inline

build.fsx
```diff
@@ -102,22 +107,32 @@ let run exe arg dir =
     let result = Process.execWithResult (psi exe arg dir) TimeSpan.MaxValue
     if not result.OK then (failwithf "`%s %s` failed: %A" exe arg result.Errors)
 
+type Communication =
+    | Remoting
+    | Bridge
+    | NoCommunication
```
theimowski: let's use `option` type and `None` for default

theimowski:
Added couple comments

Content/src/Client/Client.fs
Nhowka: This one is preventing the else branch

Content/src/Client/Client.fs
Nhowka: I'll give it a try

Content/src/Client/Client.fs
Nhowka: Oops 😅

Content/.template.config/template.json
Nhowka: Sorry, sent it on a limited internet connection and has kinda of hurried.

build.fsx
Nhowka: Huh, another example of how taking less time thinking creates overly complex things. That will even help on the other steps, good catch

Nhowka:
I was taken by surprise and got moved from workplaces where I don't have internet access yet. But I'll try to send commits using my phone as soon as possible.

theimowski:
Take your time, no need to hurry :)

Content/src/Client/Client.fs
```diff
@@ -138,11 +150,26 @@ let stream model msgs =
 let update (msg : Msg) (currentModel : Model) : Model =
     match currentModel.Counter, msg with
     | Some counter, Increment ->
+    #if (bridge)
+        Bridge.Send (ServerMsg.Increment)
```
Zaid-Ajaj: Combination of `reaction && bridge` are using the side-effects directly instead of proper commands using `Cmd.bridgeSendOr`

Nhowka:
Feels ready... Any pointers on what can be better?

theimowski:
I'll come to this next week

Content/src/Client/Client.fs
theimowski: I see, let's change the `else` branch to `elseif (!bridge)` instead - those empty lines are misleading

Content/src/Client/Client.fs
```diff
@@ -159,10 +163,40 @@ let update (msg : Msg) (currentModel : Model) : Model * Cmd<Msg> =
     | _, InitialCountLoaded initialCount->
         let nextModel = { Counter = Some initialCount }
         nextModel, Cmd.none
-
     | _ -> currentModel, Cmd.none
 #endif
 
+#if (reaction)
+
```
theimowski: those will result in many empty lines in generated project, let's try to minimise them

Content/src/Server/ServerGiraffe.fs
```diff
@@ -34,12 +38,36 @@ let port =
     "SERVER_PORT"
 //#endif
     |> tryGetEnv |> Option.map uint16 |> Option.defaultValue 8085us
+#if (bridge)
+/// Elmish init function with a channel for sending client messages
+/// Returns a new state and commands
+let init clientDispatch () =
+    let value = {Value = 42}
```
theimowski: Formatting: let's keep a single space at beginning and end of  record construction `{ Value = 42 }` - this comment applies to other places as well

build.fsx
```diff
@@ -129,18 +144,23 @@ type ServerPaketDependency = Saturn | Giraffe | Suave
...
+                    | Bridge -> "bridge")
+                |> Option.defaultValue "nocommunication"
             let azure = if x.Azure then "azure" else "noazure"
-            sprintf "%s-%s-%s" server remoting azure
+            sprintf "%s-%s-%s" server communication azure
 
 type CombinedPaketDependencies =
     { Azure : bool
-      Remoting : bool
+      Communication : Communication option
```
theimowski: Have you tried running `UpdatePaketLockFiles` after this change? We need to make sure it still works

theimowski:
This is looking good now 👍  Added couple minor comments
Can you also add some docs to the other repository?

build.fsx
Nhowka: I believe I did, I had to change some lines so the files that weren't created wouldn't fail the build target, then I reverted after creating everything and updated the files.

Content/src/Client/Client.fs
Nhowka: Nice solution! Will do

Nhowka:
Sure! I'll fix the points raised. I hope I can write a full guide soon using the template as starting point.

theimowski:
Thank you very much for this! Released in v1.8

On a side note, as per this [issue](https://github.com/SAFE-Stack/SAFE-template/pull/270) from now on we will strive to simplify the template, so we'll need to discuss every new feature before adding it

PR #270, opened 2019-06-06
no need to restore client

forki:
I don't think this is still needed.
I think I added it long time ago in bookstorish project because fable was a dotnet tool that we needed to restore explicitly.


theimowski:
thanks!

PR #271, opened 2019-06-11
Remove suave

isaacabraham:
This PR removes Suave from the SAFE template. This will:

* Make it easier document and drive the overall project in the future
* Make it easier for people to contribute to the project
* Remove the amount of code that needs to be maintained
* Remove the amount of code that needs to be tested
* Improve build time


forki:
Wow that's a lot

Isaac Abraham <notifications@github.com> schrieb am Di., 11. Juni 2019,
23:24:

isaacabraham:
> Wow that's a lot

Oh yes. Simplifying Fulma options will also help - and removing that option completely will help even more.

theimowski:
I just pushed v2 branch, let's base all breaking changes like this on that branch. I don't think we should do that as part of v1 template

isaacabraham:
OK - in that case we should create a milestone and create what changes we're going to make to the template before moving onto v2 (the sooner the better IMHO - implementing this and the Fulma simplifications would seem to be pre-requisites for the bigger changes around the componentisation).

isaacabraham:
@theimowski that being said - how should I fix the PR then - can someone help?

theimowski:
Easiest would be just to create a new PR

Issue #289, opened 2019-07-06
Feature Request: Build the Docker Container "self-contained" via docker build


rabejens:
The `--deploy docker` option does building of the Docker container a bit in a roundabout way: It compiles the application on the local machine and then copies the result into a Docker container.

However, the better practice is to have a `Dockerfile` that automatically sets up the build environment during the `docker build` process. So, one who just wants to build the application as a Docker container without wanting to install all the prerequisites just has to type `docker build -t whatever .`

Since both approaches have their advantages and disadvantages, I suggest an additional option `--deploy docker-build`.

I already made the changes necessary, I'll post a PR on this shortly.

theimowski:
Actually we already have an option of building the app in container using VS Code Remote containers extension, see here for details: https://github.com/SAFE-Stack/SAFE-template/pull/254

theimowski:
Also, because of #269 we need to restrain from adding new options like suggested

sebastian:
I just wanted to add to @rabejens original point: this is not necessarily only for local development. In that case using VS Code Remote might be a decent option. It is particularly interesting when you want to build your containers on a third party host, particularly if the host is not under your control. Be it a CI server or for example a [dokku](http://dokku.viewdocs.io/dokku/) host (or any other PaaS solution that supports deployments from Dockerfiles). In these cases you likely do not want to pollute the server with the necessary dependencies even in the cases where you are able to do so.

olivercoad:
Agreed, I'd absolutely expect to be able to build the container using docker build.

Rather than adding complexity with an additional `--deploy` option, is there any reason not to just change the existing `--deploy docker`  option to do the build in docker?

theimowski:
Fixed by https://github.com/SAFE-Stack/docs/pull/179

PR #290, opened 2019-07-06
Have the Docker build process set up its environment

rabejens:
This PR adds a `--deploy docker-build` option which sets up the `Dockerfile` and `build.fsx` in a way the Docker container can just be built by going into the root directory and typing `docker build -t mytag .`

This solves #289


theimowski:
Closing as v2 does no longer support `--deploy` option

PR #291, opened 2019-07-09
Add .ionide to gitignore

bruinbrown:
Ionide builds up a symbol cache which it stores locally, we don't need this in source control. Given that we encourage people to use Ionide in VS we should ensure people don't encounter any unnecessary friction.


theimowski:
thanks!

PR #297, opened 2019-07-19
Fix runTool working directory

rmunn:
Instead of specifying the `runTool` working directory as `"."`, which might be any directory depending on where the build script was run from, we should use `__SOURCE_DIRECTORY__` which will always be consistent.


theimowski:
thanks!

PR #298, opened 2019-07-19
Support for emitting test results as NUnit XML.

isaacabraham:
This will make it possible to include test results in devops pipelines and reports.


theimowski:
Shouldn't this result in https://compositional-it.visualstudio.com/SAFE%20Template/_build/results?buildId=3208&view=ms.vss-test-web.build-test-results-tab showing passed test?

isaacabraham:
It does now - it didn't because the build itself didn't have the Test task in it. [This build](https://compositional-it.visualstudio.com/SAFE%20Template/_build/results?buildId=3209&view=ms.vss-test-web.build-test-results-tab) was manually kicked off by me and does have the test task in the build, and therefore does work.

Unfortunately Expecto shows it as a single test rather than multiple (one for each alternative version of FSCheck's inputs) - I'm not sure if that can be changed though?

theimowski:
I don't think you can show result for every input - not sure if that made sense since it's generated in a random fashion, so you couldn't compare runs

PR #299, opened 2019-07-19
Remove reaction option

### LINKED ISSUE #272: Remove obsolete "reaction" pattern option
The `reaction` option for the `pattern` argument is obsolete (having been replaced with `streams`) and should be removed.

Also consider renaming `default` with something like `commands` or `elmish` or `basic` instead?

theprash:
This fixes #272.


theimowski:
thanks :+1:

PR #302, opened 2019-08-08
Fix indentation for DU definition and pattern match

theprash:
Fix indentation to be in line with [F# code formatting guidelines](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting).


theimowski:
Have we checked rest of code adheres to that rule as well? Best to be consistent across all parts

theprash:
Yep I looked through all the other code files and they already do it "right" 😀

theimowski:
thanks :+1:

PR #305, opened 2019-08-11
Set font by customizing a Bulma variable

kerams:
Now that Bulma is part of the bundle, it's possible to customize any of its variables before we import it, so there's no need for CSS in this particular case.


kerams:
It might be useful to also have this link https://bulma.io/documentation/customize/variables/ in a comment there

kerams:
Is it the newline that breaks the build or what?

theimowski:
probably - I've seen that before, not sure why this happens. Can you revert whitespace changes?

kerams:
Looks good now.

theimowski:
thanks!

PR #311, opened 2019-08-27
Add AppName environment parameter to create app without "safe-" prefix and "-web" suffix

JonathanDDuncan:
I wasn't happy with how prescriptive the ARM template and Build Script were. Because they always prefixed the website name with "safe-" and suffixed it with "-web". So I made some changes so that I can call my site whatever I want. I've also added a proper appName environment variable to set the app Service Name.

I also pulled some of the variables out of the ARM template and made them into parameters so that the names can be cleaned in the script before passing them to the ARM template. This will make cleaning the names easier in the future for anybody who wants to customize names. It is now checking for an acceptable name for storage which didn't always work before if the name "environment" was too long. I also checks that the storage name is lowercase alfanumeric.  I've also added filtering for the resourceGroup name with has to be alfanumeric or _-.()

It is still possible to omit the appName and Environment variables and get the GUID name as before.

### DIFF
--- a/Content/arm-template.json
+++ b/Content/arm-template.json
@@ -2,7 +2,16 @@
     "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
     "contentVersion": "1.0.0.0",
     "parameters": {
-        "environment": {
+        "fullAppName": {
+            "type": "string"
+        },
+        "storageName": {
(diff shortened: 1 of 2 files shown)


isaacabraham:
Sorry - I've literally just seen this PR now! I'll review it this week. Apologies for the delay.

PR #339, opened 2020-03-06
Use FAKE and Paket local tools in template output

theprash:
This removes FAKE and Paket as a global dependencies, using dotnet local tools instead.

- Tests don't work for me locally before or after my changes, possibly due to environmental issues.
- I've not tested the Dockerfile. I don't really use docker.

I haven't moved away from global tools for the SAFE-template project itself, but we should do that as the next step, unless @theimowski is doing this for v2 anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,9 +24,6 @@ cache:
   directories:
   - $HOME/.yarn-cache
 
-env:
-  - FAKE_PATH=~/fake-cli/fake MAX_TESTS=10
-
 script:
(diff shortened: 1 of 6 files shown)


theprash:
The instructions in the docs repo needs to be updated too.

If [dotnet tool restore is made automatic](https://github.com/dotnet/sdk/issues/4241) then we can eventually remove a step from the instructions.

theimowski:
Thanks, this is looking good - Travis is currently broken but Azure pipelines look ok.

PR #340, opened 2020-03-09
Update publisher for vscode csharp extension

olivercoad:
The extension for csharp has changed publisher from ms-vscode to ms-dotnettools 

https://github.com/OmniSharp/omnisharp-vscode/pull/3610


theimowski:
thanks!

PR #341, opened 2020-03-10
Fix devServerProxy context paths

olivercoad:
In the dev server proxy, it proxies requests from `/api/*` to the server.
This PR fixes it so that it only proxies requests starting with `/api/` instead of all starting with `/api`.
See https://github.com/webpack/webpack-dev-server/issues/2454 for details


theimowski:
thanks!

PR #352, opened 2020-04-27
Set working directory for debugging server

### LINKED ISSUE #239: Consider setting the CurrentDirectory in the template code
Often when an application is run under the ionide debugger, the `System.Environment.CurrentDirectory` is set to that of the debugging instance, which is somewhere inside the user home. This causes relative-path file lookups, like a bundled sqlite database for example, to fail. Should the template include code to set the `System.Environment.CurrentDirectory` to the location of the currently-executing assembly(`System.Reflection.Assembly.GetExecutingAssembly().Location`)?  This was inspired by a troubleshooting session I had this morning with a Slack user in #web.

olivercoad:
This sets the working directory in the vscode launch config for debugging the server.

Fixes #239

### DIFF
--- a/Content/.vscode/launch.json
+++ b/Content/.vscode/launch.json
@@ -12,6 +12,7 @@
                 "moduleLoad": false
             },
             "stopAtEntry": false,
+            "cwd": "${workspaceFolder}/src/Server",
             "console": "internalConsole",
         },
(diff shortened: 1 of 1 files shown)


theimowski:
thanks!

PR #353, opened 2020-04-28
update Thoth.Fetch version

TheDevKnight:
Hi,

i had some issues with simple calls to a REST service because the template use the old Thoth.Fetch version 1.1.0 for the Client

https://github.com/MangelMaxime helped me to analyse my problem https://github.com/thoth-org/Thoth.Fetch/issues/16

and the solution was to bump up the version of his lib. It would be good to update the version in the template as well, so that other new devs will not have my problems :)

Regards


theimowski:
Closing in favour of #356

PR #355, opened 2020-05-09
Upgrate thoth fetch to version 2

robertpi:


isaacabraham:
Does this not need all the lock files updated as well? @theimowski 

Also - not sure if at this stage we will make anything other than support hotfixes to the v1 template given that v2 is right around the corner.

isaacabraham:
But we should definitely use the work you've done here in the v2 template!

theimowski:
Yup, it does require updating lock files (`fake build -t UpdatePaketLockFiles -- Client`)
I'll submit a separate PR that does it. 
#353 is related

theimowski:
Closing in favour of #356

PR #358, opened 2020-05-19
global.json: SDK version "3.1.*"

theprash:
Without this change the template didn't seem to install correctly when trying to test it locally if I have a slightly more recent SDK version installed.


theimowski:
I'd rather pin that version - feel free to point to the latest SDK there

isaacabraham:
@theimowski I think moving to a "min version" should be fine, no? This ease the onboarding process as well.

theimowski:
Just read about [rollForward](https://docs.microsoft.com/en-us/dotnet/core/tools/global-json?tabs=netcore3x#rollforward) setting - how about using the `latestFeature` option + version of `3.1.300`? 
If I understand correctly that would mean `>= 3.1.300` and `< 3.2`

isaacabraham:
I would suggest we go with min version is 3.1 and no max. We can then just bump that minimum as we take on dependencies on new specific features of e.g. net5 etc.

theimowski:
When there was an update from .net core 2.1 to 3.0, there were breaking changes for SAFE template. Same will probably happen for next major update. But I can understand the onboarding ease concern - how about we go with `3.1.300` + `latestMinor` ?

theprash:
This has been solved elsewhere 😀

PR #360, opened 2020-05-27
V2 minimal

isaacabraham:
@theimowski @theprash @MangelMaxime I've gone through the template based on the [this repo](https://github.com/isaacabraham/safe-lite) and added it all into the minimal template. Can any of you review (@Zaid-Ajaj this is as we discussed)?

Main changes:

* Simplified the server a lot (some of the nonsense that was needed for Azure and such is actually not needed anymore).
* Client.fs is now a separate file - this I think makes it more readable and gives us more scope to make changes being the minimal and full versions.
* Specific readme for Minimal.
* GitIgnore based on a larger file from VSCode's GitIgnore -> VisualStudio template.
* Minor changes to webpack and server for static content paths - changed the output folder for production builds - this now means that you can copy the entire folder directly into wherever you publish the server and it will "just work".

If possible (and I know this is hard as this is a somewhat subjective topic) can we keep the changes to a minimum? Keep in mind:

* This is not about documenting and learning any more - we will have documentation that will support that coming.
* This is about a non-opinionated, bare-bones template version that can be used for either
    * SAFE experts that want a "hand crafted" approach to building their app.
    * Non-SAFE developers that want to experience the basic SAFE experience without any bells and whistles or extra technology that may confuse / complicate the experience.

I'll be moving onto the "full" version tomorrow and have it complete by Friday.


theimowski:
@isaacabraham  FYI I've been doing some more changes to v2_minimal branch, but haven't pushed them here - here's the PR https://github.com/SAFE-Stack/SAFE-template/pull/361 but not insisting on getting everything from there. Anyway feedback would be welcome

Content/.gitignore
```diff
@@ -1,16 +1,362 @@
-.fable/
-.fake/
+## Ignore Visual Studio temporary files, build results, and
```
theimowski, line 1: As this is the "minimal" option, I'd keep everything as minimal as possible - including gitignore
As an "expert" wanting to have everything in control I'd not be happy with a full-blown gitignore file

Content/src/Client/Client.fsproj
```diff
@@ -9,8 +9,7 @@
     <None Include="package.json" />
     <None Include="package-lock.json" />
     <None Include="webpack.config.js" />
-<!--#if (minimal)-->
```
theimowski, line 12: please keep those as they are - will be helpful when switching from minimal to full and back while working on the template

Content/global.json
```diff
@@ -1,5 +1,6 @@
 {
   "sdk": {
-    "version": "3.1.100"
+    "version": "3.1.100",
+    "rollForward": "latestMajor"
```
theimowski, line 4: as discussed in the other PR, I'd opt for `latestMinor`

Content/src/Client/client-minimal.fs
```diff
@@ -0,0 +1,73 @@
+module Client
```
theimowski, line 1: Extracting a separate file like that makes it impossible to work on the template for minimal version (it won't compile)

Content/src/Server/Server.fsproj
```diff
@@ -5,8 +5,7 @@
     <TargetFramework>netcoreapp3.1</TargetFramework>
   </PropertyGroup>
   <ItemGroup>
-<!--#if (minimal)-->
```
theimowski, line 8: as above - let's leave those

Content/.gitignore
isaacabraham: That's a good point. Perhaps just create two different gitignore files?

Content/src/Client/client-minimal.fs
isaacabraham: Do you work on the template directly? Sorry - I thought that that wasn't even possible. We just worked off a separate repository, got it working and then copied the files over into here.

theimowski: It wasn't possible with v1, but it is possible with just the full and minimal options. I'll write a short instruction on how to do that in CONTRIBUTING.md

Content/.gitignore
theimowski: Not sure if there's a need to have the auto-generated gitignore, even for full template - any specific reasons there?

isaacabraham: I assumed that a standard one will catch most elements that may crop up in the future once people start working on a project and not need to add to the file. Happy to revert.

theprash: For VS I think you only need `.vs/` these days. I am against having these uber gitignores.

I might even go so far as to say it's up to the user to manage this in their user-level gitignore because they could be using any IDE, and we should only ignore things inherent to the project like `node_modules` etc. But I'm OK with covering VS, VS Code, Ionide and Rider which should only be 4 lines.

isaacabraham: There are surely some others as well - bin and obj etc. etc.?

theprash: I was only referring to the ones that come from your choice of IDE/editor.

Was there a problem with the existing gitignore? It had 16 lines and covered everything we needed. It would work perfectly for the full/opinionated template option too.

MangelMaxime:
Hello @isaacabraham I just read the code in the PR and it seems good to me.

isaacabraham:
@theimowski I think that this part is ready for the "minimal" branch. I've undone the changes of the conditions and the gitignore file.

isaacabraham:
If we can merge this in, I will look at the working off your branch this weekend.

PR #361, opened 2020-05-27
More changes to V2 minimal

theimowski:


isaacabraham:
@theimowski I thought we weren't going to introduce testing until post-v2 - especially for the minimal. That "lite" repo that we've been working on for the past week or so was (I thought) the ideal place to experiment before investing the time into modifying the template directly.

isaacabraham:
@theimowski saw your comment on the other repo - sorry.

My personal view is that we should resist the temptation to add more features now - I would rather get the basic template out using the current counter model as a v2 and then put in a simple set of aims for v2.1, such as moving to this to-do list, adding testing support, looking at feliz etc.

theimowski:
Requests for adding tests have been there for some time already, so I'd like to include tests for full template option (they're almost ready). 

I'm not convinced about the need to hurry to release v2 just to bring a lot other changes shortly afterwards with v2.1 - but let me think the list through and arrive at some compromise

isaacabraham:
@theimowski should this be closed or stay open?

theimowski:
yup closing

PR #364, opened 2020-06-01
Fix devServerProxy context paths in V2 minimal

olivercoad:
Noticed that #341 didn't make it into one of the webpack configs in v2_minimal


theimowski:
correct - the branch was created a while ago and some of the recent master changes haven't been yet pulled in. We should revisit recent changes to v1 before polishing v2 @isaacabraham

PR #365, opened 2020-06-03
V2 tests

theimowski:
and changing counter to todo list
@isaacabraham


theimowski:
@isaacabraham @Krzysztof-Cieslak @theprash @dustinmoris

I've checked into this PR the changes we discussed:
1) Use todo list sample for full option; use "minimal" sample for minimal option (even removed the counter thing)
2) Renamed back the project and module / namespaces names to just `Shared`, `Server` and `Client`
3) Still having the issue with Shared fsproj sometimes not being recognised as a dependency of Client project, not sure why.

@Zaid-Ajaj I'd be grateful if you could review the way tests are configured in the full template option - most of it is taken from your examples.

isaacabraham:
I'll review in the next day or so.

isaacabraham:
@theimowski just checked, the minimal looks fine to me. I'm happy to merge this in and go live and iterate on it for 2.1 etc.

theimowski:
Thanks, did you check the full option with tests as well?

isaacabraham:
Nope - I can do that as well though. @Zaid-Ajaj do you want to review this as well?

Zaid-Ajaj:
LGTM, though a couple of points:
 - Usually `npm test` is not a long-running process but rather something you can run once and get test results, maybe rename it to `npm run test:live` so that later on you can add `npm test` that does the testing all at once (to run in CI servers)
 - The webpack config for the test project should look as close as possible to the config of the front-end itself (the `Client.fsproj`), ideally sharing common configuration pieces like loaders and plugins and only adjusting the project/output path. So that things will work ootb when for example importing images, CSS files or including polyfills in your code. Otherwise the test code will break in your tests

theimowski:
Thanks @Zaid-Ajaj I applied your suggestions. 
The only outstanding issue is that `Shared.fsproj` or `Shared.Tests.fsproj` sometimes is not seen by Fable - anyone aware of a known issue here?

isaacabraham:
@theimowski do we need a dedicated "shared.tests" project or can we not just put that in the server tests project?

isaacabraham:
@theimowski reliability has IMHO to be a must-have for the template; if the shared project stuff doesn't work every time, we either should wait until Fable fixes this, or revert to the shared file approach. What do you think?

isaacabraham:
@theimowski what do you think about changing the default FAKE target to Run rather than Bundle? Most people are using SAFE from a dev perspective so this could just be "fake build"?

isaacabraham:
@theimowski aside from my questions above, the test part works great - we just need to document it all up (we knew this anyway). Can I merge this PR into v2-minimal? We can iterate on it from there with a view to releasing soon.

theimowski:
Yup I'm happy to get this merged. I'm a bit hesitant about changing default FAKE target, for a reason similar to Zaid's comment around NPM script - IMO people usually expect the default build task (whatever the project is and what ecosystem it uses) to simply compile whole project rather than entering a "watch" mode

theimowski:
I think I found the issue - seems to be a race condition on compiling Shared project in `run` FAKE target - both Server and Client try to compile it at the same time and if Server does it faster then Client gets compilation errors. My latest commit to this PR does build Shared project first, before compiling Server and Client in parallel - tried couple of times and after that change the issue seems to be gone.

isaacabraham:
Travis is failing because of the specific NET Core required. Can we merge #358 in (and then change to use max minor) to fix this?

theimowski:
we can but TBH I think Travis could fail anyway - I've recently ignored its failures

isaacabraham:
I'm going to merge this today unless any reasons why not?

I would like to release on Friday.

theimowski:
:shipit:

isaacabraham:
There's that squirrel again.

PR #369, opened 2020-06-19
Simplify FAKE script a bit

theprash:
We don't need to do: `if Environment.isUnix then "npm" else "npm.cmd"`

### DIFF
--- a/Content/build.fsx
+++ b/Content/build.fsx
@@ -16,17 +16,12 @@ let deployDir = Path.getFullName "./deploy"
 
 let npm args workingDir =
     let npmPath =
-        let npmPath = if Environment.isUnix then "npm" else "npm.cmd"
-
-        npmPath
-        |> ProcessUtils.tryFindFileOnPath
-        |> Option.defaultWith(fun () ->
-            let errorMsg =
-                npmPath +
-                " was not found in path. " +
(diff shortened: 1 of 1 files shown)


Content/build.fsx
```diff
@@ -16,17 +16,12 @@ let deployDir = Path.getFullName "./deploy"
...
-        |> Option.defaultWith(fun () ->
-            let errorMsg =
-                npmPath +
-                " was not found in path. " +
-                "Please install it and make sure it's available from your path. " +
-                "See https://safe-stack.github.io/docs/quickstart/#install-pre-requisites for more info"
-            failwith errorMsg)
+        match ProcessUtils.tryFindFileOnPath "npm" with
+        | Some path -> path
+        | None ->
```
isaacabraham, line 21: Could this be simplified into `Option.defaultWith(fun () -> ...)`?

theprash: I actually just changed it away from that because I thought this was a bit simpler!

isaacabraham: True. I find that the `Some x -> x | None -> exn` pattern is more easily encoded with `defaultWith` - you can see a little more easily what's happening IMHO. But not fussed.

PR #373, opened 2020-07-04
V2 SAFE

isaacabraham:
This PR merges in the V2 Minimal template into master. @theimowski this (unsurprisingly) doesn't merge in cleanly. I would be quite happy to just ignore everything in master and override with what's in v2_minimal. What do you think?


theimowski:
we shouldn't just force merge it - there have been some significant changes to the master branch that we might have omitted from v2 branch. So I'll need to take some time to identify those changes and apply them somehow

isaacabraham:
I had a (very quick!) look through but everything I saw seemed to be made obsolete by v2. Might be wrong though.

isaacabraham:
I suspect you could probably merge from master into this branch and just let this branch "win" all conflicts?

isaacabraham:
Regarding docs - we have started a few recipes but I think best is to get the new template out and then we can spend time rapidly moving the docs around. I have time on Friday to work on this as do the rest of our team here.

olivercoad:
> I suspect you could probably merge from master into this branch and just let this branch "win" all conflicts?

This looks like a use case for the **ours** merge strategy (not to be confused with the ours options for the recursive merge strategy).

I think you could do it with

```bash
git checkout v2_minimal
git merge -s ours master
```

isaacabraham:
@theimowski just noticed that font awesome doesn't work with the default template. Either we need to add the CSS link to the HTML or add the NPM package.

HTML add:

```html
<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.14.0/css/all.min.css">
```

isaacabraham:
Fixed the issue with FA.

isaacabraham:
@theimowski I've noticed that client tests seem to be broken (when launching on 8081) with the error "Failed to load resource: the server responded with a status of 404 (Not Found)" (for Client.Tests.js) and then an error "Target container is not a DOM element." for react-dom-development.

I have no clue how this works, I'm afraid - any chance you can have a look?

isaacabraham:
I've now merged upwards from master into this branch - it was pretty smooth in the end and I think aside from the issue with the client-side testing this is very, very close to being ready!

PR #374, opened 2020-07-10
Use Program.withReactSynchronous to avoid input focus bug

theprash:
Whenever we add more than a handful of input fields to a SAFE project we see a bug where any input field change makes the cursor jump to the end of the field. Using `Program.withReactSynchronous` instead of `Program.withReactBatched` fixes this.


isaacabraham:
Thanks. This is a common issue I've seen people fall foul of.

PR #376, opened 2020-07-23
Try fix tests

isaacabraham:


theimowski:
note that `dotnet` is invoked from Path, not fully qualified one - so it might potentially take a version of dotnet where you have a different safe template installed

theimowski:
should work now :+1:

isaacabraham:
@theimowski Thanks dude. Have you tests it? I can't (windows here) - if it works locally feel free to merge.

PR #383, opened 2020-08-13
The redundant cssEntry deleted from webpack.config.js

functionalprogrammer:


theimowski:
@isaacabraham wanted to bring back sass loader support to full template, maybe this entry will be relevant afterwards?

isaacabraham:
@functionalprogrammer if we get the other PR in that you created, can we kill this PR off or should we keep it?

isaacabraham:
This isn't needed I think - closing.

PR #386, opened 2020-08-20
Simplify run test

isaacabraham:
This combined the Run and RunTests fake targets in a single parameterised function. It also uses `--no-restore` to speed up performance of .NET Watch.

### DIFF
--- a/Content/build.fsx
+++ b/Content/build.fsx
@@ -63,23 +63,16 @@ Target.create "Azure" (fun _ ->
     |> ignore
 )
 
-Target.create "Run" (fun _ ->
+let run clientCommand sharedPath serverPath _ =
     dotnet "build" sharedPath
-    [ async { dotnet "watch run" serverPath }
-      async { npm "run start" clientPath } ]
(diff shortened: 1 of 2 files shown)


isaacabraham:
I accidentally created this on top of the Farmer PR. Sorry.

Content/build.fsx
```diff
@@ -63,23 +63,16 @@ Target.create "Azure" (fun _ ->
     |> ignore
 )
 
-Target.create "Run" (fun _ ->
+let run clientCommand sharedPath serverPath _ =
```
theimowski, line 66: not convinced by this change - yes it makes script shorter, but adds another layer of complexity

isaacabraham: I must admit I'm not 100% sure of it either. I did it because I wanted to add the --no-restore for both. But now I've realised that you still need it because the build command doesn't run against the Server project, it runs against Shared. Which means that when you do a watch run against the server, it fails.

So maybe just bin this one.

PR #390, opened 2020-08-25
Move npm and webpack files to root

### LINKED ISSUE #389: v2 - Fable compile error when adding any DU to shared code
1. dotnet new safe
2. Add any DU to the Shared.fs file, e.g. `type X = A | B`
3. dotnet fake build -t run

This results in a Fable compile error:

```
ERROR in ../Shared/Shared.fs
Module not found: Error: Can't resolve 'core-js/modules/es.function.name' in 'C:\temp\safe-template-bug\src\Shared'
 @ ../Shared/Shared.fs 1:0-42
 @ ./Index.fs
 @ ./App.fs
 @ ./Client.fsproj
```

I have node v14.5.0.

I have found that updating core-js from 3.3.2 to 3.6.5 (current latest) fixes the issue.

theprash:
This fixes #389. I've tested this with both configurations, running the app, running tests and bundling.


Content/src/Client/Client.fsproj
```diff
@@ -19,17 +15,17 @@
...
-    <PackageReference Include="Fable.Core" Version="3.1.5" />
-    <PackageReference Include="Fable.Elmish" Version="3.0.6" />
-    <PackageReference Include="Fable.Elmish.React" Version="3.0.1" />
-    <PackageReference Include="Fable.Elmish.Debugger" Version="3.2.0" />
-    <PackageReference Include="Fable.Elmish.Hmr" Version="4.0.1" />
-    <PackageReference Include="Fable.React" Version="5.3.6" />
-    <PackageReference Include="Thoth.Fetch" Version="2.0.0" />
-  </ItemGroup>
-  #else-->
+    <!--#if (minimal)-->
```
isaacabraham, line 18: Is this just indent changes or has the extra --> been added?

isaacabraham (approved):
Looks good aside from that one query.

Content/src/Client/Client.fsproj
theprash: The extra XML comment syntax just makes it so that you can edit the template file without it breaking syntax highlighting in your editor. This is in line with what was already being done further up in the file.

The only change to this part of the file with the PackageReferences in the output is correcting the indentation.

theimowski:
I can remember I've put the webpack files in that Client directory for a reason. Too bad I can't remember what the reason was 😞  But if all run configurations (including) tests are fine then we should probably be fine

PR #394, opened 2020-08-28
Use `do` instead of `->` when yielding todos

theprash:
I think we should avoid the arrow syntax as many people aren't aware that it exists and it doesn't play nicely with implicit yield. If you put another item adjacent to it you have to use the yield keyword.


forki (approved):
+

theimowski:
makes sense - thanks!

PR #400, opened 2020-09-09
Add loading indicator to the model and 'add' button.

ChrSteinert:
The first thing I did when playing with the template, was adding the loading indication to the button. I really like the feel of it and I think it adds some depth and adds to the understanding of how to use those things.

### DIFF
--- a/Content/src/Client/Index.fs
+++ b/Content/src/Client/Index.fs
@@ -42,6 +42,7 @@ open Shared
 
 type Model =
     { Todos: Todo list
+      IsLoading : bool
       Input: string }
 
 type Msg =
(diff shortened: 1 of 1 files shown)


theimowski (commented):
I like that change. Maybe to make it even more "real-worldy" we could:
* mark input as disabled when model isloading
* add a small `Async.Sleep` server side to simulate latency
* clean input upon `AddedTodo` rather than `AddTodo`

The only drawback is it makes the example code a bit more complex. Curious to hear @isaacabraham  thoughts on that

isaacabraham:
I'm in favour of adding this, but I would refer also to #402 which I think addresses a bigger issue.

isaacabraham:
Someone has already started creating a more fully-featured todo implementation [here](https://github.com/davedawkins/SAFE-template-extended). I wonder whether we could pull this in as an official demonstrator app and focus our energy on getting that as good as possible?

ChrSteinert:
Let me close this for now, until a course of action has been chosen :)

PR #401, opened 2020-09-10
Separate dirs in template codebase to maintain minimal and default options

theimowski:
This should make it seamless to work on template and test it directly from template repository directory


isaacabraham (approved):
LGTM. Much easier to see what's going on - especially as I've seen people starting to go directly to this repository to see how code should look for examples.

PR #415, opened 2020-12-07
Net5.0 & fable3

theimowski:


Zaid-Ajaj:
[Fable.Remoting#201 - Fable.Remoting.Giraffe broken in Giraffe 5](https://github.com/Zaid-Ajaj/Fable.Remoting/issues/201) is a blocker for this upgrade where latest Giraffe v5.0-pre that works with net5 isn't compatible with the Giraffe adapter from Fable.Remoting. It's a bit tricky on remoting side of things because upgrading the adapter also break backward-compat with applications using Giraffe [>= 3.6 && < 5]

theimowski:
Thanks for the heads up. Until that issue is solved I'll pin Giraffe version to below v5

theimowski:
@theprash do you think it's a good idea to move away from paket-based dependencies for fake build script and instead use the `#r "nuget: Package"` syntax? Wondering if that's the recommended way for fake scripts - can we guarantee pinned versions of resolved packages?

Zaid-Ajaj:
> Thanks for the heads up. Until that issue is solved I'll pin Giraffe version to below v5

@theimowski I just pushed a new version of Fable.Remoting.Giraffe v5.0.0-rc-6 that uses latest prerelease of Giraffe targeting `net5`. It should work in principle, let me know if you encounter any issues when using it

isaacabraham:
@theimowski we've tried and it doesn't work for some reason that I don't fully understand (but Prash does). I think for now we've had to run our FAKE scripts via fsi or something.

theprash:
@theimowski The main reason for moving was that FAKE seems to be tied to its own paket version which didn't support .NET 5. However, I never really liked the fake-cli approach, though I'm sure it has some advantages. One problem is that `#r "paket: groupref build //"` is a compile error everywhere except VS Code as far as I'm aware. And you need the additional 
`#load "./.fake/build.fsx/intellisense.fsx"`.

RE package resolution, no I don't think you can have completely reproducible package resolution with this method if NuGet doesn't provide it.

Another option is to run FAKE as a console project instead. I quite like this approach because you can easily use paket and you don't need any special tooling like fake-cli or any special `#r` commands that IDEs don't understand. You just `dotnet run` the console project. You can pass through the command line arguments and set up the FAKE environment in the same or similar way to what I already added to the top of `build.fsx`. It might be worth add a function to FAKE itself that can do this in one line.

theimowski:
Yeah I think @Zaid-Ajaj also likes the idea of invoking FAKE via standard fsproj rather than fsx. I'll experiment with that approach.
The `#r nuget` syntax is great for quick setup and fire-and-forget scripts. However as it doesn't guarantee locked resolution, I'd rather not user it to run FAKE.

theimowski:
Need to park updating to latest Giraffe prerelease, as Saturn needs to catch up https://github.com/SaturnFramework/Saturn/issues/281

theimowski:
af5bc09 is a PoC of using FAKE in fsproj rather than as fsx script

theimowski:
Updated the PR to latest prerelease of Saturn and Giraffe - seems to work nice locally. 
Would be curious to hear your thoughts on a different approach for running build as opposed to build.fsx - see updated README.md for default template.
Still need to fix the tests.

.config/dotnet-tools.json
```diff
@@ -9,7 +9,7 @@
       ]
```
theprash, line 9: I think you can remove `fake-cli` from this file now?

Content/default/Build.fs
```diff
@@ -1,20 +1,18 @@
...
-
 open Fake.Core
 open Fake.DotNet
 open Fake.IO
 open Farmer
 open Farmer.Builders
 
-Target.initEnvironment ()
+let execContext = Context.FakeExecutionContext.Create false "build.fsx" [ ]
+Context.setExecutionContext (Context.RuntimeContext.Fake execContext)
```
theprash, line 8: I think we should add a FAKE function to do this so we can just write something like `Target.initExeEnvironment ()`. Or maybe even remove the need for this if possible when using FAKE this way.

Content/default/Build.fs
```diff
@@ -1,20 +1,18 @@
...
-#r "netstandard"
-
 open Fake.Core
 open Fake.DotNet
 open Fake.IO
 open Farmer
 open Farmer.Builders
 
-Target.initEnvironment ()
+let execContext = Context.FakeExecutionContext.Create false "build.fsx" [ ]
```
theprash, line 7: We're passing an empty list for `args` here. I'm surprised that this works correctly to pick a target. Does passing in variables also work? e.g. `dotnet run -- e Setting=value`

theprash:
I think it looks pretty good. Having a Build.fsproj means we have an extra file in the root directory, but it means that we can remove `fake-cli` and stick with the .NET build system that's already being used (and the same version of paket). And we also get the benefits I described [above](https://github.com/SAFE-Stack/SAFE-template/pull/415#issuecomment-742478736). I think it's an inherently simpler and more understandable approach.

One downside is the new way to pass arguments to the FAKE script: `dotnet run -- RunTests`. The double-dash looks a little bit odd but at least it's standard .NET tooling thing. And overall we have the same number of "words" as before in the command.

Another problem is that we need all this code in `Build.fs`:

```fsharp
let execContext = Context.FakeExecutionContext.Create false "build.fsx" [ ]
Context.setExecutionContext (Context.RuntimeContext.Fake execContext)
```

I'm not even sure this is correct at the moment for passing in all arguments, but I think FAKE should easily support this scenario without needing to write this code manually. (CC: @matthid 👋🏾)

.config/dotnet-tools.json
theimowski: these are dotnet tools for building the template itself - currently still using fake tool here.
I've removed fake tool from the default template though.

Content/default/Build.fs
theimowski: Not sure - I don't really use passing in variables like this, I usually rely on environment variables

Content/default/Build.fs
theimowski: Yup it is a boilerplate, but I could live with those two extra lines

theimowski:
@isaacabraham how do you configure the Azure Pipelines build? Do you need to do that from within visualstudio online?

isaacabraham:
@theimowski what sort of configuration are you referring to? You can more or less do all devops stuff from YAML, given even time (and patience).

theimowski:
The thing is I can't see any YAML file in the repository, for travis we have `.travis.yml` but Azure Pipelines is a black box for me

Content/default/.config/dotnet-tools.json
```diff
@@ -3,15 +3,15 @@
...
-      "version": "6.0.0-alpha014",
+      "version": "6.0.0-beta8",
       "commands": [
         "paket"
       ]
     },
-    "fake-cli": {
-      "version": "5.19.1",
+    "fable": {
+      "version": "3.0.0",
```
schauerte, line 12: Would it be a problem to go straight to the current version?
```suggestion
      "version": "3.1.2",
```

theimowski:
This is now ready to be reviewed, though we might wait for stable releases of Giraffe and Saturn 
@isaacabraham @theprash @Zaid-Ajaj

Content/default/README.md
```diff
@@ -1,14 +1,16 @@
...
 
-* The [.NET Core SDK](https://www.microsoft.com/net/download) 3.1 or higher.
-* [npm](https://nodejs.org/en/download/) package manager.
-* [Node LTS](https://nodejs.org/en/download/).
+* [.NET Core SDK](https://www.microsoft.com/net/download) 5.0 or higher
+* [Node LTS](https://nodejs.org/en/download/)
 
 ## Starting the application
+
 Before you run the project **for the first time only** you must install dotnet "local tools" with this command:
```
theprash, line 14: Now that we don't use a dotnet tool to run the build, we can remove this instruction and put `dotnet tool restore` in the FAKE code instead. Another advantage of this approach.

theimowski: I think it's better to continue asking to do `dotnet tool restore` - otherwise you might not have paket installed and restore can fail

bengobeil:
Hi, what is the state on this PR? Any known reason for timeout? Anything I can add to this to accelerate the process?

isaacabraham:
@bengobeil I think @theprash has been looking into this with regards to the FAKE integration / issues with dotnet 5. I wouldn't be averse to releasing this as a beta (are they supporting for dotnet templates yet?).

Also note that there are still some Azure issues with .NET 5 - App Service support is experimental and code-free App Insights integration through the App Service doesn't work at all.

bengobeil:
Beta would be cool. Just something people can get their hands on and at least get the dividends of @theimowski and @theprash 's work. Starting a new project and would be cool to have Fable 3 and .NET 5, even if there are hiccups, I can deal with it.

bengobeil:
On the FAKE side of things, their seems to be issues with maintainer (I empathize). But there is a PR there ready to go, to better support .NET 5.

bengobeil:
Maybe a post-mortem would be cool, like how can we better coordinate with the releases of .NET for key community-driven projects that are invaluable to the language. Maybe F# software foundation should support key projects in the ecosystem to make sure the development experience meets acceptable standards. It seems to be a recurring problem.

isaacabraham:
We have a working internal project with both - perhaps we could strip out the code and leave a bare bones sample for you to look at in the meantime? @theprash

isaacabraham:
There is a SAFE Stack working group - we try to meet up every 6 weeks or so to coordinate on SAFE related stuff but for .NET in general, that's a bigger issue to be sure.

theimowski:
@isaacabraham @bengobeil This PR is ready for review. There is a couple of caveats:
* Giraffe and Saturn are still prerelease versions
* The PR suggests a new way of using FAKE - inside a standard dotnet project rather than fsx script. In my opinion this works pretty well and is inspired by @Zaid-Ajaj idea
* The CI tests fail in non-deterministic fashion - sometimes the execution hangs, hence the timeout on Travis. I don't know how to configure Azure pipelines though as I don't have access to the account @isaacabraham

theimowski:
Travis looks better now - seems that for some unkown reason Expecto runner would sometimes hang the execution. I'm now running the tests without expecto runner. So now just azure pipelines need to be fixed.

isaacabraham:
I'll dig into the Azure pipeline.

isaacabraham:
One observation by moving to a console app for the build is that starting your SAFE app is now just `dotnet run` (I notice that the default target is now `run` rather than `build` as well).

isaacabraham:
I would like to use this opportunity to minimise some of the output - especially from Fable - if possible. @alfonsogarciacaro is there a "silent" flag for Fable - something that wouldn't emit console logging for every single `.fs` file that was compiled?

isaacabraham:
@theimowski the problem with the DevOps agent is that `dotnet build` doesn't work (see [here](https://compositional-it.visualstudio.com/SAFE%20Template/_build/results?buildId=7343&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=33e5d3ec-87e7-5f80-0281-074c6962cb44)) any longer.

The DevOps pipeline is well out of date and needs to be refreshed - we should move over to the code-first approach like Travis. I'll see if one of the team can fix this in the next days.

isaacabraham:
The Azure DevOps build pipeline has now been fixed.

alfonsogarciacaro:
@isaacabraham There used to be a `--silent` flag but I just realized Fable 3 ignores it 😅 Actually I was thinking of the same, maybe we can use the same line for the "Compiled Foo.fs" messages, what do you think?

screencast

bengobeil:
Great job everyone. Can't wait to get started.

isaacabraham:
@alfonsogarciacaro Yep, that would help - I imagine that we would use that for the default "local" build / run but for a CI/CD you would probably turn that off so you get the full output. Or would that "same line" log everything to e.g. Travis / DevOps console anyway? If so - maybe consider making this "single line" approach the default? :-)

olivercoad:
> Travis looks better now - seems that for some unkown reason Expecto runner would sometimes hang the execution. I'm now running the tests without expecto runner. So now just azure pipelines need to be fixed.

Are you sure it was expecto runner that was the cause? I tried checkout out 605ab4a and changing it to run without the expecto runner, and it still hangs, suggesting that it was somthing else.

9b96119 also changes the tests from running `dotnet new SAFE` in a temp dir to running directly in the content directories.

alfonsogarciacaro:
@isaacabraham Yes, my thinking was to make the one-line log for compiled files the default. Looking at Appveyor and Travis in the Fable build, the logs don't seem to show all the lines but this is actually not a super useful log, it was meant mainly so the user could see some kind of progress (particularly before when Fable was slower). We could even change it to some kind of progress indicator.

theimowski:
@olivercoad I'm not 100% sure what was the root cause of hanging and timeouts, but when I switched to invoking the test directly it started working 🤷

olivercoad:
@theimowski I think it probably had more to do with making it run directly in the content folders that made it start working.

theimowski:
And why would that be?

theimowski:
Prerelease version of template now available with `dotnet new -i SAFE.Template::3.0.0-beta001`

theprash:
@theimowski CONTRIBUTING.md still says you should run `dotnet fake build` to build the template. If I do `dotnet run` I get "Specify which project file to use because C:\Users\Prash\dev\SAFE-template contains more than one project file". It only works with the longer command `dotnet run -p Build.fsproj`.

Do you know why it doesn't work in the same way as the output SAFE app?

bengobeil:
Are you referring to the Template repo are the generated SAFE template? In the template repo, there are two proj's (Build.fsproj and SAFE.Template.proj). When you generate a template, `dotnet run` works as expected.

theprash:
@bengobeil The template repo. Ah right I didn't notice that other proj file. So `dotnet run -p Build.fsproj` is what we need to update CONTRIBUTING.md then 👍🏾

bengobeil:
Has anyone noticed that `dotnet paket add -p <path> <package_id>` no longer works? It adds the dependency to the root paket.references file and does not add it to the denoted project.

ddombrow:
Not sure if this is due to my setup but I get an error trying to use the beta template under WSL2, has to do with the paket project path that gets inferred I think:
```dotnet paket restore --project "\\wsl$\Ubuntu-18.04\[snip]\Metro\src\Shared\Shar
ed.fsproj" --output-path "obj" --target-framework "net5.0"
Paket version 6.0.0-beta8
Total time taken: 0 milliseconds
Paket failed with
-> DirectoryNotFoundException: Could not find a part of the path '/wsl$/Ubuntu-18.04/[snip]/Metro/src/Shared/Shared.fsproj'.```

**Comment by isaacabraham** [comment]
Closes #424

**Comment by olivercoad** [comment]
@ddombrow The problem is it's finding and using the dotnet.exe on windows instead of using the dotnet installed in WSL.

I've opened an issue https://github.com/fsharp/FAKE/issues/2573

A workaround is to override DotNetCliPath in the fake dotnet exec options

```fsharp
// around L35 in build.fs
let dotnet cmd workingDir =
    let dotnetPath =
        Option.defaultValue "dotnet" (ProcessUtils.tryFindFileOnPath "dotnet")

    let result =
        DotNet.exec
            (DotNet.Options.withWorkingDirectory workingDir
             >> DotNet.Options.withDotNetCliPath dotnetPath)
            cmd
            ""
```

theimowski:
3.0.0-beta002 includes now Fantomas formatting plus a couple small fixes

mastoj:
Tried the latest beta from linux now and `dotnet run` fails.

```
Fake.Core.BuildFailedException: Target 'Run' failed.
 ---> System.AggregateException: One or more errors occurred. (Start of process 'dotnet' failed.)
 ---> System.Exception: Start of process 'dotnet' failed.
 ---> System.ComponentModel.Win32Exception (2): No such file or directory
   at System.Diagnostics.Process.ForkAndExecProcess(String filename, String[] argv, String[] envp, String cwd, Boolean redirectStdin, Boolean redirectStdout, Boolean redirectStderr, Boolean setCredentials, UInt32 userId, UInt32 groupId, UInt32[] groups, Int32& stdinFd, Int32& stdoutFd, Int32& stderrFd, Boolean usesTerminal, Boolean throwOnNoExec)
   at System.Diagnostics.Process.StartCore(ProcessStartInfo startInfo)
   at System.Diagnostics.Process.Start()
   at Fake.Core.Process.processStarter@415.Invoke(RawCreateProcess c, Process p)
   --- End of inner exception stack trace ---
   at Fake.Core.Process.processStarter@415.Invoke(RawCreateProcess c, Process p)
   at Fake.Core.RawProc.start@207(FSharpFunc`2 startProcessRaw, RawCreateProcess c, ProcessStartInfo p, StreamSpecs streamSpec, Process toolProcess, CancellationTokenSource tok, FSharpRef`1 isStarted, FSharpRef`1 readOutputTask, FSharpRef`1 readErrorTask, FSharpRef`1 redirectStdInTask, Unit unitVar0)
   at Fake.Core.RawProc.syncStart@268.Invoke(Unit unitVar0)
--- End of stack trace from previous location ---
   at Microsoft.FSharp.Control.AsyncResult`1.Commit() in D:\workspace\_work\1\s\src\fsharp\FSharp.Core\async.fs:line 337
   at Microsoft.FSharp.Control.AsyncPrimitives.RunSynchronouslyInCurrentThread[a](CancellationToken cancellationToken, FSharpAsync`1 computation) in D:\workspace\_work\1\s\src\fsharp\FSharp.Core\async.fs:line 870
   at Microsoft.FSharp.Control.AsyncPrimitives.RunSynchronously[T](CancellationToken cancellationToken, FSharpAsync`1 computation, FSharpOption`1 timeout) in D:\workspace\_work\1\s\src\fsharp\FSharp.Core\async.fs:line 878
   at Microsoft.FSharp.Control.FSharpAsync.RunSynchronously[T](FSharpAsync`1 computation, FSharpOption`1 timeout, FSharpOption`1 cancellationToken) in D:\workspace\_work\1\s\src\fsharp\FSharp.Core\async.fs:line 1142
   at Fake.Core.Proc.run[a](CreateProcess`1 c)
   at Fake.DotNet.DotNet.exec(FSharpFunc`2 buildOptions, String command, String args)
   at Build.dotnet(String cmd, String workingDir) in /home/tomas/git/mastoj/simpleshopper/Build.fs:line 36
   at Build.clo@65-4.Invoke(TargetParameter _arg5) in /home/tomas/git/mastoj/simpleshopper/Build.fs:line 66
   at Fake.Core.TargetModule.runSimpleInternal(TargetContext context, Target target)
   --- End of inner exception stack trace ---
   --- End of inner exception stack trace ---
   at Fake.Core.TargetModule.raiseIfError(OptionalTargetContext context)
   at Fake.Core.TargetModule.runOrDefault(String defaultTarget)
   at Build.main(String[] args) in /home/tomas/git/mastoj/simpleshopper/Build.fs:line 109
```

`dotnet --version` gives `5.0.103`.

theimowski:
@mastoj thanks for reporting - this is quite strange, tried to reproduce it on dotnet 5 sdk docker image, but it works fine 🤷  
Could be something related to `DotNet.exec` - I think I'm gonna change it do invoke `dotnet` directly since we now include `global.json` in the template

mastoj:
@theimowski , no problem. Let me know if I can do anything more figuring out what's wrong. 

Here is my OS information if it is to any help:

```
➜ cat /etc/os-release
NAME="Pop!_OS"
VERSION="20.10"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 20.10"
VERSION_ID="20.10"
HOME_URL="https://pop.system76.com"
SUPPORT_URL="https://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=groovy
UBUNTU_CODENAME=groovy
```

theimowski:
Try replacing `dotnet` function in `Build.fs` with following please:

```fsharp
let run exe arg dir =
    CreateProcess.fromRawCommandLine exe arg
    |> CreateProcess.withWorkingDirectory dir
    |> CreateProcess.ensureExitCode
    |> Proc.run
    |> ignore

let dotnet cmd workingDir =
    run "dotnet" cmd workingDir
```

mastoj:
Hm, not sure what happened with my environment. All of sudden the dotnet SDK I had installed was downgraded, I did play around with apt-get and apt maybe that caused it. After that I couldn't find the SAFE template locally any more and installing again didn't help. I'll update here when things are running again.

mastoj:
I won't be able to reproduce it exactly the same since it looks like I changed something in my environment. As of now I can't install any templates unless  I do a `sudo` install, which seems odd.

theimowski:
FWIW the templates are tied to dotnet SDK version installed, so if you have multiple versions of dotnet SDK installed then the list of templates may vary - e.g. if you cd to directory with `global.json` file it's gonna forward `dotnet` commands to use the pinned version

theimowski:
I have released beta003 - added `Helpers.fs` to Build project to hide a few helper functions for the build logic. 
Most notably, I added a helper `Proc.Parallel.run` that can take 2 process definitions, run them in parallel but take extra care of not interleaving the outputs of both. This is pretty much inspired by npm package called [concurrently](https://www.npmjs.com/package/concurrently)
Looking forward to hear feedback about these - if people like it I plan to extract them to Fake PR.

isaacabraham:
@theimowski what's your thoughts on current status - do you think we can start to think about scheduling a release of this?

theimowski:
From my PoV the outstanding issues to be addressed are:
1. hear feedback about what I did for [beta003](https://github.com/SAFE-Stack/SAFE-template/pull/415#issuecomment-814432374) and whether we want to include it for v3
2. whether we shall wait for stable release of [Giraffe on .net 5](https://github.com/giraffe-fsharp/Giraffe/issues/471)
3. how we deal with docs / recipes updates

Other issues (but IMO not critical) include:
* whether the new fsproj-based Build works ok
* how to pass arguments to Build targets

isaacabraham:
Linking to [docs issue](https://github.com/SAFE-Stack/docs/issues/201)

Content/default/Helpers.fs
```diff
@@ -0,0 +1,95 @@
+module Helpers
+
+open Fake.Core
+
+let initializeContext () =
+    let execContext = Context.FakeExecutionContext.Create false "build.fsx" [ ]
+    Context.setExecutionContext (Context.RuntimeContext.Fake execContext)
+
+module Proc =
+    module Parrallel =
```
theprash, line 10: Typo. Should be `Parallel`.

Akash-Mair:
Hey @theimowski 

I've just tried running this latest version of the SAFE Template and I'm getting the following error when I run `dotnet run`

```
Fake.Core.BuildFailedException: Target 'InstallClient' failed.                                                           ---> System.AggregateException: One or more errors occurred. (Start of process 'npm' failed.)                           ---> System.Exception: Start of process 'npm' failed.                                                                   ---> System.ComponentModel.Win32Exception (2): The system cannot find the file specified.                                 at System.Diagnostics.Process.StartWithCreateProcess(ProcessStartInfo startInfo)                                        at System.Diagnostics.Process.Start()                                                                                   at Fake.Core.Process.processStarter@415.Invoke(RawCreateProcess c, Process p)                                           --- End of inner exception stack trace ---                                                                              at Fake.Core.Process.processStarter@415.Invoke(RawCreateProcess c, Process p)                                           at Fake.Core.RawProc.start@207(FSharpFunc`2 startProcessRaw, RawCreateProcess c, ProcessStartInfo p, StreamSpecs streamSpec, Process toolProcess, CancellationTokenSource tok, FSharpRef`1 isStarted, FSharpRef`1 readOutputTask, FSharpRef`1 readErrorTask, FSharpRef`1 redirectStdInTask, Unit unitVar0)                                                              at Fake.Core.RawProc.syncStart@268.Invoke(Unit unitVar0)                                                             --- End of stack trace from previous location ---                                                                          at Microsoft.FSharp.Control.AsyncResult`1.Commit() in D:\workspace\_work\1\s\src\fsharp\FSharp.Core\async.fs:line 337                                                                                                                           at Microsoft.FSharp.Control.AsyncPrimitives.RunSynchronouslyInCurrentThread[a](CancellationToken cancellationToken, FSharpAsync`1 computation) in D:\workspace\_work\1\s\src\fsharp\FSharp.Core\async.fs:line 870                               at Microsoft.FSharp.Control.AsyncPrimitives.RunSynchronously[T](CancellationToken cancellationToken, FSharpAsync`1 computation, FSharpOption`1 timeout) in D:\workspace\_work\1\s\src\fsharp\FSharp.Core\async.fs:line 878                      at Microsoft.FSharp.Control.FSharpAsync.RunSynchronously[T](FSharpAsync`1 computation, FSharpOption`1 timeout, FSharpOption`1 cancellationToken) in D:\workspace\_work\1\s\src\fsharp\FSharp.Core\async.fs:line 1142                            at Fake.Core.Proc.run[a](CreateProcess`1 c)                                                                             at Helpers.run[a,b,c](FSharpFunc`2 proc, a arg, b dir) in C:\Users\Akash\Documents\code\delete\Helpers.fs:line 78       at Build.clo@20-1.Invoke(TargetParameter _arg2) in C:\Users\Akash\Documents\code\delete\Build.fs:line 20                at Fake.Core.TargetModule.runSimpleInternal(TargetContext context, Target target)                                       --- End of inner exception stack trace ---                                                                              --- End of inner exception stack trace ---                                                                              at Fake.Core.TargetModule.raiseIfError(OptionalTargetContext context)                                                   at Fake.Core.TargetModule.runOrDefault(String defaultTarget)                                                            at Helpers.runOrDefault(String[] args) in C:\Users\Akash\Documents\code\delete\Helpers.fs:line 91
```

But if I manually run `npm install` at the root and comment out the `InstallClient` step then `dotnet run` the app works.

theimowski:
Thanks @Akash-Mair for reporting - that's probably because I removed [this line](https://github.com/SAFE-Stack/SAFE-template/pull/415/commits/f3d46a49476c6bc5d18d4f6d1c5cbead60059969#diff-bfed9727842a66e1d0b857f5817bf2f0f622ba2d9d7f4afe3e6a7f8eb1cb86c1L19) as I wanted to simplify the build. But I'll need to bring it back as indeed on windows my simplification might not always work

theimowski:
should be fixed in beta004

Content/default/SAFE.App.sln
```diff
@@ -31,6 +31,8 @@ Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "Server.Tests", "tests\Serve
 EndProject
 Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "Client.Tests", "tests\Client\Client.Tests.fsproj", "{CF8C20AC-1FC4-4114-ACB8-6CE6E51F34F9}"
 EndProject
+Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "Build", "Build.fsproj", "{C0244434-EC34-4D74-93A9-0E00DEC0C4D1}"
```
FredericEspiau, line 34: Line 8, `build.fsx` doesn't exist anymore

olivercoad: That's `Build.fsproj` not `build.fsx`

FredericEspiau: I was referring to the line 8 of this file that I can't comment on: https://github.com/SAFE-Stack/SAFE-template/pull/415/files/0251d8cdf2b26618d86b66320283c87564e2f7ba#diff-f564918bd20243543596c1f4ec459f93c958aa325432596895e431f9a15f126aL8

As this was done: https://github.com/SAFE-Stack/SAFE-template/pull/415/files/0251d8cdf2b26618d86b66320283c87564e2f7ba?file-filters%5B%5D=.fsx#diff-a26cc857ceabaa57f1ac087600547fd4cfd4db7c28c641d264378d2d5de4f294L1

Or maybe I understood something wrong

olivercoad: Yes you're right. Can't comment on L8 due to https://github.com/isaacs/github/issues/1655

theprash: I've addressed this here: https://github.com/SAFE-Stack/SAFE-template/pull/443

FredericEspiau: This has been fixed

theimowski:
Do we still want to keep Babel in the default template? We can get rid of some additional packages and webpack config plus get less npm vulnerabilities in future by doing so, the drawback is loosing compat with IE - as Alfonso mentions [here](https://github.com/MangelMaxime/fulma-demo/pull/43#issuecomment-715631483)

FredericEspiau:
> Do we still want to keep Babel in the default template? We can get rid of some additional packages and webpack config plus get less npm vulnerabilities in future by doing so, the drawback is loosing compat with IE - as Alfonso mentions [here](https://github.com/MangelMaxime/fulma-demo/pull/43#issuecomment-715631483) 
> 

People who want IE compatibility can add it themselves, it should not be the default behavior in 2021 imho

theimowski:
Giraffe is now on stable v5.0, the only outstanding prerelease versions we depend in this PR are following (pinging maintainers to ask if we can expect a stable release soon?):
* Paket 6.0.0-rc001 @forki 
* Saturn 0.15.0-preview03 @Krzysztof-Cieslak 
* Fable.Remoting.Giraffe 5.7.0-rc-6 @Zaid-Ajaj

isaacabraham:
@theimowski there's another blocker we've noticed recently which is related to Fable Remoting (we think) that's causing issues on Azure - it's forcing you to run in 64 bit mode which isn't available on the free tier, which I really want to keep.

isaacabraham:
@theimowski I think removing Babel is fine as long as we have a recipe for adding support for IE

kerams:
> Fable.Remoting.Giraffe 5.7.0-rc-6

v5 of this package is in lockstep with Giraffe itself, so now that the latter is stable, there's no technical reason that I know of why Fable.Remoting.Giraffe should not follow suit.

@isaacabraham, how do you figure? By process of elimination? Please open an issue if you know more.

isaacabraham:
@kerams there's already some open issue on Fable.Remoting (I think!) - I think it's related to another library though - one by Eirik Tsarpalis (though I can't remember the name). @Akash-Mair can you let us know where that issue is / was?

kerams:
It's https://github.com/Zaid-Ajaj/Fable.Remoting/issues/226, but I took it that the problem had been resolved. [The new report](https://github.com/SAFE-Stack/SAFE-template/issues/440) is based on the current stable version of the template, right? If that's the case, I believe it could be referring to an old version of Remoting without the fix, so we should (hopefully) be good here.

Zaid-Ajaj:
@theimowski A stable package of Fable.Remoting.Giraffe v5.0 has been published that depends on Giraffe v5.0

There were also bug fixes to Fable.Remoting.Client which can also be updated to latest v7.12.0 (char type serialization / ignoring connectivity errors when status code = 0 of HTTP requests)

isaacabraham:
@theimowski @Zaid-Ajaj there's a PR open that upgrades all deps to latest.

isaacabraham:
Wow. I think we're good to do - there's a conflict on the project-local json that I think can be overwritten by what's in this branch?

FredericEspiau:
> Wow. I think we're good to do - there's a conflict on the project-local json that I think can be overwritten by what's in this branch?

https://github.com/SAFE-Stack/SAFE-template/blob/master/Content/minimal/package.json

Makes sense yes, the conflict shouldn't matter much and the latest changes by Dependabot are already taken into account:
- https://github.com/SAFE-Stack/SAFE-template/pull/433

Dependabot latest PR

- https://github.com/SAFE-Stack/SAFE-template/blob/1b28e1e79c39681b73447cd60b8eaac783ab46e5/Content/minimal/package-lock.json#L10026

`elliptic` is already at version `6.5.4` in the current PR

theimowski:
Yeah the conflict can probably be even resolved locally if necessary - just force merge using this branch changes as it diverged quite significantly from master anyway.
I'm happy to roll out stable v3 release as soon as we agree docs are ready too 👍

isaacabraham:
@theimowski right, everything looks ready. let's do it!

comintel:
Beta 4 works for me, but beta 5 and 3.0.0 give this error:

PS D:\data-vs-code-workspaces\SAFEDir> dotnet new SAFE --force
Template "SAFE-Stack Web App v3.0.0" could not be created.
Error while processing file /Content/default/node_modules.orig/nan/nan.h
System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
   at System.Collections.Generic.List`1.get_Item(Int32 index)
   at Microsoft.TemplateEngine.Core.Expressions.Cpp.CppStyleEvaluatorDefinition.Evaluate(IProcessorState processor, Int32& bufferLength, Int32& currentBufferPosition, Boolean& faulted)
   at Microsoft.TemplateEngine.Core.Operations.Conditional.Impl.HandleMatch(IProcessorState processor, Int32 bufferLength, Int32& currentBufferPosition, Int32 token, Stream target)
   at Microsoft.TemplateEngine.Core.Util.ProcessorState.Run()
   at Microsoft.TemplateEngine.Core.Util.Processor.Run(Stream source, Stream target, Int32 bufferSize, Int32 flushThreshold)
   at Microsoft.TemplateEngine.Core.Util.Processor.Run(Stream source, Stream target, Int32 bufferSize)
   at Microsoft.TemplateEngine.Core.Util.Processor.Run(Stream source, Stream target)
   at Microsoft.TemplateEngine.Core.Util.Orchestrator.ProcessFile(IFile sourceFile, String sourceRel, String targetDir, IGlobalRunSpec spec, IProcessor fallback, IEnumerable`1 fileGlobProcessors, IReadOnlyList`1 locOperations)

davidhayesbc:
I see the same issue as comintel

theimowski:
My bad - for some reason I had a dirty directory locally that got packaged into the nupkg. Should be solved now with 3.0.1

davidhayesbc:
Thanks, that fixed it for me! Looking forward to getting started on my F# web journey

comintel:
Working perfectly now for me also,  THANK YOU

PR #418, opened 2021-01-03
Fix vulnerabilities

dedale:
- ini: cherry picked dependabot's commits
- object-path: seems not supported by dependabot. Manually updated resolve-url-loader


theimowski:
thanks!

Issue #219, opened 2019-01-25
Use FAKE packages for yarn / npm related tasks


theimowski:
NuGets:
* Fake.JavaScript.Yarn
* Fake.JavaScript.NPM

Need to check if all used functionality is in the packages

PR #426, opened 2021-02-22
Make Tests work on windows

olivercoad:
There were two issues preventing tests working on windows.

 The first is `pgrep`, used for getting child process to kill the process tree. Fixed using `wmic` when on windows.
The second is the path to `npm`. The fix code to locate it is copied from fake until we address #219

### DIFF
--- a/tests/Tests.fs
+++ b/tests/Tests.fs
@@ -17,6 +17,17 @@ let dotnet =
     match Environment.GetEnvironmentVariable "DOTNET_PATH" with
     | null -> "dotnet"
     | x -> x
+
+let npm =
+    // copied from Fake.JavaScript.Npm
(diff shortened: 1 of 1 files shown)


theimowski:
Great thanks! Would be awesome to get these running on Travis Windows matrix, but last time I checked Travis C# didn't support windows

olivercoad:
Don't really have experience with travis so not sure how the language support works. Could windows install dotnet 5 with chocolatey maybe?

theimowski:
We're using [community C# travis support](https://docs.travis-ci.com/user/languages/csharp/) and it seems it doesn't work with windows out of the box. Not sure how much effort is required to set something like this up.

PR #428, opened 2021-02-26
Format with Fantomas

nojaf:
Hello, 

Apologies for not opening an issue first, it might be easier to talk about this when seeing the impact.
So, in short, I'd like to add Fantomas to this template in order to gain some more exposure.

Fantomas follows the [Microsoft F# style guide](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting) by default so that would be the main selling point of having it here.

I could add some more things (FAKE targets, tweaks in configuration, add it to the minimal template, ...) but I wanted to hear first if you are interested in any of this.

Thank you for your time.


Content/default/src/Client/Index.fs
```diff
@@ -15,29 +13,33 @@ type Msg =
     | AddedTodo of Todo
 
 let todosApi =
-    Remoting.createApi()
+    Remoting.createApi ()
```
kerams, line 16: Big fan of this :)

Content/default/src/Client/Index.fs
```diff
@@ -15,29 +13,33 @@ type Msg =
...
-    | SetInput value ->
-        { model with Input = value }, Cmd.none
+    | GotTodos todos -> { model with Todos = todos }, Cmd.none
+    | SetInput value -> { model with Input = value }, Cmd.none
     | AddTodo ->
         let todo = Todo.create model.Input
-        let cmd = Cmd.OfAsync.perform todosApi.addTodo todo AddedTodo
+
+        let cmd =
+            Cmd.OfAsync.perform todosApi.addTodo todo AddedTodo
```
kerams, line 36: Just curious, surely the line wasn't considered too long?

nojaf: That part is controlled by the max_line_length setting.

theimowski:
Hi @nojaf - thanks for this. Yes I'd definitely like to see Fantomas in the template! 
Will review the changes in a bit.
General question: I know there's a VS Code setting for Ionide that invokes Fantomas formatting when you save the file - can we add it to the PR as well? Also what about other editors - VS, Rider - do they have a setting like this?

Content/default/.config/dotnet-tools.json
```diff
@@ -13,6 +13,12 @@
       "commands": [
         "fable"
       ]
+    },
+    "fantomas-tool": {
```
theimowski, line 17: Should there also be a corresponding FAKE target to format the sources?
What's the preferred way of referencing Fantomas from FAKE - via dotnet tool or NuGet package?

Content/default/.editorconfig
```diff
@@ -5,4 +5,10 @@ indent_style = space
 indent_size = 4
 charset = utf-8
 trim_trailing_whitespace = true
-insert_final_newline = false
\ No newline at end of file
+insert_final_newline = false
+
+[*.fs]
+fsharp_space_before_colon=true
```
theimowski, line 11: This is not the default option, right? 
If not then why include it - personal preference?
In general I believe we should start with default configuration

Content/default/src/Client/Index.fs
theimowski: And the default now is what? 120? 80?

Content/default/tests/Client/Client.Tests.fs
```diff
@@ -5,27 +5,29 @@ open Fable.Mocha
...
+              Expect.equal 1 model.Todos.Length "There should be 1 todo"
+              Expect.equal newTodo model.Todos.[0] "Todo should equal new todo" ]
 
 let all =
-    testList "All"
+    testList
+        "All"
         [
-//-:cnd:noEmit
+          //-:cnd:noEmit
```
theimowski, line 25: not sure if this extra space here won't fool the template engine

Content/default/tests/Client/Client.Tests.fs
```diff
@@ -5,27 +5,29 @@ open Fable.Mocha
...
 
-let client = testList "Client" [
-    testCase "Added todo" <| fun _ ->
-        let newTodo = Todo.create "new todo"
-        let model, _ = init ()
+let client =
+    testList
+        "Client"
+        [ testCase "Added todo"
+          <| fun _ ->
```
theimowski, line 12: this looks a bit weird. In fact I'm not a fan of the `<|` operator anyway. Maybe you can suggest a different syntax here?

Content/default/.editorconfig
```diff
@@ -5,4 +5,10 @@ indent_style = space
...
+
+[src/Client/Index.fs]
+fsharp_single_argument_web_mode=true
```
theimowski, line 14: This is for the Feliz formatting, isn't it? If so we should probably extend it to all fs files under `src/Client`

theimowski (commented):
Looking good! Couple of comments / questions

Content/default/.config/dotnet-tools.json
nojaf: The best approach of late is to use the .NET tool via FAKE.
For example:
https://github.com/fsprojects/fantomas/blob/ed7a94071cc7deeba849ec7806d67801094b37e6/build.fsx#L376-L409

I can add those targets if you like.

Content/default/.editorconfig
nojaf: This was to match the existing style I found. Is not the default indeed.
Remove?

Content/default/.editorconfig
nojaf: Indeed, this is to have Feliz style. If you have files that do nothing with Feliz this could potentially lead to unexpected results, not in line with the formatting guide. That is why I pinned down the exact file.

Content/default/src/Client/Index.fs
nojaf: Default is 120, [see default settings](https://github.com/fsprojects/fantomas/blob/master/docs/Documentation.md#configuration).

Content/default/tests/Client/Client.Tests.fs
nojaf: Hmm, perhaps to be on the safe side (pun intended), the test files [should be ignored](https://github.com/fsprojects/fantomas/blob/master/docs/Documentation.md#ignore-files-fantomasignore).

nojaf:
When it comes to editors, the story is a bit more complex.
There is short cut in Rider `Reformat Code` and I believe there would be a similar thing in Visual Studio (@deviousasti any ideas?)
However, these will format the code with a version of Fantomas that is bundled with the editor.

In Ionide this is a version of Fantomas found on [NuGet](https://github.com/fsharp/FsAutoComplete/blob/f1b1a5e4813e1b1f8bf79139c260ee17c6f040a6/paket.dependencies#L14).
Visual Studio is using the source of Fantomas and this is tied closely to the [master branch or latest stable](https://github.com/deviousasti/fsharp-formatting-for-vs).
Rider uses [a fork of Fantomas](https://github.com/auduchinok/fantomas) (that is only updated when they release).

So in terms of formatting, the FAKE targets might be a more interesting route to explore. Because there the repository controls the Fantomas version.

Content/default/tests/Client/Client.Tests.fs
theimowski: Just tested it - indeed the engine doesn't like it, this is what gets produced:
```fsharp
let all =
    testList
        "All"
        [
          #if FABLE_COMPILER // This preprocessor directive makes editor happy
          Shared.Tests.shared
#endif
                    client ]
```

Yeah I'm thinking maybe start small by formatting only `src/` files and leave `tests/` for now?

theimowski:
Understood - I'm happy to get this in when review comments get addressed. I'd also like to get a 👍 from @isaacabraham @theprash @Krzysztof-Cieslak

Content/default/.config/dotnet-tools.json
nojaf: Yes, I have given that some thought and I will pick this up someday as part of the next major version.
No real timeline for that yet though.

Content/default/src/Client/Index.fs
nojaf: Oh, there are two things in play here:
* the max_line_length will determine if the function application should remain in one line or go multiline:
```fsharp
Cmd.OfAsync.perform 
    todosApi.addTodo 
    todo 
    AddedTodo
```
* setting [fsharp_max_value_binding_width](https://github.com/fsprojects/fantomas/blob/master/docs/Documentation.md#fsharp_max_value_binding_width) controls wether the expression after the `=` sign remains on the same line. (Threshold is 40 there)

Content/default/tests/Client/Client.Tests.fs
nojaf: Good idea!

deviousasti:
Hey @nojaf, as you've rightly pointed out, FantomasVS deploys both versions - stable and bleeding edge which can be selected from. 

There's also a format on save feature, and can be enabled from `Options > F# Tools > Formatting`.

I'd like to support a feature where if a Fantomas version is specified in dotnet's `global.json` tools, to use that if possible. That way there's likely to be less of a mismatch with formatting tests in the build pipeline.

theimowski:
This is great, thanks again @nojaf

PR #434, opened 2021-03-12
Pinned TypeShape version to 8.0.1 to fix deployment issue

Akash-Mair:
Running an azure deployment would break since a different version of TypeShape was being used (9.0.0) 

For more details - https://github.com/Zaid-Ajaj/Fable.Remoting/issues/226


Content/default/paket.lock
```diff
@@ -381,8 +381,9 @@ NUGET
     Thoth.Json (5.1)
       Fable.Core (>= 3.1.6)
       FSharp.Core (>= 4.7.2)
-    TypeShape (9.0)
-      FSharp.Core (>= 4.3.2)
+    TypeShape (8.0.1)
```
eiriktsarpalis, line 384: I would recommend avoiding taking a nuget dependency and embed the source file directly. You can then pin the version of the library you depend on at compile time.

https://github.com/eiriktsarpalis/TypeShape#installing

theimowski:
a08bd5b updates Fable.Remoting.Server to latest and removes transitive TypeShape dependency (see https://github.com/Zaid-Ajaj/Fable.Remoting/commit/42bfab7f208e979fb55a1fe6b643d930913fe3f2)

isaacabraham:
I think we can close this PR without merging?

Zaid-Ajaj:
Yeah Fable.Remoting no longer requires TypeShape as a nuget dependency

PR #436, opened 2021-03-14
Fixed link for Saturn documentation

SCullman:
https://saturnframework.org/docs returns a 404

### DIFF
--- a/Content/default/README.md
+++ b/Content/default/README.md
@@ -36,6 +36,6 @@ If you want to know more about the full Azure Stack and all of it's components (
 
 You will find more documentation about the used F# components at the following places:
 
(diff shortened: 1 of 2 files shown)


theimowski:
Thanks!

theimowski:
actually I'll use the root path to saturn website

SCullman:
@theimowski, how about swapping out the [Elmish](https://elmish.github.io/elmish/) link with the [Elmish Book](https://zaid-ajaj.github.io/the-elmish-book/#/)? The users can learn much more there.

theimowski:
I think we can add Elmish Book as a second reference - but I'd like to keep the original website link

PR #443, opened 2021-05-07
Add dotnet fable clean

theprash:
While working on SAFE apps I often find that Fable gets into a bad state with cached JS files which isn't fixed by restarting the build script, so I always add this command to delete all `*.fs.js` files.


theimowski:
thanks :+1:

theimowski:
@theprash  this change has one unwanted consequence: when running `dotnet run` and `dotnet run -- RunTests` in parallel, one of them will fail because the other will wipe the required files. I'm gonna remove `Clean` target frorm `RunTests` chain to mitigate it

PR #453, opened 2021-06-19
Update paket deps include Farmer.

isaacabraham:
I've run a paket update on the lock file and tested locally - it builds, creates a working app and also deployed to Azure without issues.


Content/default/paket.lock
```diff
@@ -73,26 +73,26 @@ NUGET
...
+      Fable.SimpleJson (>= 3.21)
       FSharp.Core (>= 4.7)
-    Fable.Remoting.Giraffe (5.1)
-      Fable.Remoting.Server (>= 5.21)
+    Fable.Remoting.Giraffe (5.3)
+      Fable.Remoting.Server (>= 5.23)
       FSharp.Core (>= 5.0)
       Giraffe (>= 5.0)
-    Fable.Remoting.Json (2.16)
+    Fable.Remoting.Json (2.17)
```
isaacabraham, line 86: @Zaid-Ajaj This will be the proposed lock file for v3. I think from what you've said elsewhere, this is the correct version of Remoting?

theimowski (approved):

PR #454, opened 2021-06-24
Replace deprecated Expecto entrypoint

boechat107:
This PR replaces the server test's entry point with the recommended function. 

## Why

Expecto's `runTests` is deprecated.

https://github.com/haf/expecto#runtests.


theimowski:
thanks!

PR #465, opened 2021-08-18
Rework default Server.fs - move to a module

isaacabraham:
Small reformatting but also move from a class to a module and let-bound functions.

### DIFF
--- a/Content/default/src/Server/Server.fs
+++ b/Content/default/src/Server/Server.fs
@@ -6,38 +6,31 @@ open Saturn
 
 open Shared
 
-type Storage() =
-    let todos = ResizeArray<_>()
-
-    member __.GetTodos() = List.ofSeq todos
-
-    member __.AddTodo(todo: Todo) =
+module Storage =
+    let todos = ResizeArray ()
+    let addTodo (todo: Todo) =
         if Todo.isValid todo.Description then
             todos.Add todo
             Ok()
         else
             Error "Invalid todo"
 
-let storage = Storage()
-
-storage.AddTodo(Todo.create "Create new SAFE project")
-|> ignore
-
-storage.AddTodo(Todo.create "Write your app")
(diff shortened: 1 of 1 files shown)


Content/default/src/Server/Server.fs
```diff
@@ -6,38 +6,32 @@ open Saturn
 
 open Shared
 
-type Storage() =
-    let todos = ResizeArray<_>()
-
-    member __.GetTodos() = List.ofSeq todos
-
-    member __.AddTodo(todo: Todo) =
+module Storage =
```
theimowski, line 9: it looks more succinct and maybe even more idiomatic. I'm not convinced though - prefer to use explicit class/object notation when dealing with internal state. Probably a matter of preference though - curious what other think

isaacabraham: @theimowski in absence of any feedback (happy for you to loop in some people explicitly) - I would suggest we go with this one and add a recipe for the class-based (with DI) approach.

Content/default/src/Server/Server.fs
```diff
@@ -6,38 +6,32 @@ open Saturn
...
 let todosApi =
-    { getTodos = fun () -> async { return storage.GetTodos() }
-      addTodo =
-          fun todo ->
-              async {
-                  match storage.AddTodo todo with
-                  | Ok () -> return todo
-                  | Error e -> return failwith e
-              } }
+    {
```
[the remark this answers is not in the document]
isaacabraham: Probably not.

isaacabraham:
@theimowski I'm going to merge this in on Monday unless there's any real objections.

theimowski:
Ok - feel free to do so 👍

sob., 13 lis 2021, 18:07 użytkownik Isaac Abraham ***@***.***>
napisał:

PR #466, opened 2021-08-18
Reformat default server.fs - use DI

isaacabraham:
Keeps the class-based structure, but makes the following changes:

* Move the initialisation of the in-memory DB into the Storage class.
* Registration of the service as a singleton into ASP .NET service collection.
* Retrieval of the service through the Http Context's `GetService` method.
* Small formatting changes.

### DIFF
--- a/Content/default/src/Server/Server.fs
+++ b/Content/default/src/Server/Server.fs
@@ -2,52 +2,50 @@ module Server
 
 open Fable.Remoting.Server
 open Fable.Remoting.Giraffe
+open Microsoft.AspNetCore.Http
+open Microsoft.Extensions.DependencyInjection
 open Saturn
-
 open Shared
 
-type Storage() =
-    let todos = ResizeArray<_>()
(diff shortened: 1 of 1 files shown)


olivercoad:
I'm guessing you're choosing between these two PRs. I like #465. This one would be great as a recipe for how to use DI.

isaacabraham:
@olivercoad exactly. And yes, my preference is the other one too :-)

theimowski:
not sure which one I prefer TBH :) need to give it some more thoughts

isaacabraham:
I suggest we close this and convert into a recipe / guide for DI with SAFE Stack.

isaacabraham:
Closing this - one of us will create a recipe explaining the basics of DI within the context of the server-side SAFE Stack components.

isaacabraham:
cc: @Ryan-Palmer-CIT @theprash @Akash-Mair

PR #470, opened 2021-08-20
Fable outputs to "output" folder

### LINKED ISSUE #459: Send fable output to a specific folder?
You can redirect fable to output to a specific folder e.g. `fable watch -o output --run webpack-dev-server` (note the use of `-o` here).

This redirects all outputs into a single folder - all compiled JS files, as well as the `.fable` folder. At least one of our customers has suggested that they prefer using that rather than putting the JS outputs directly into the source folder as `.fs.js` files.

I've tested this out locally and the change would be trivial:

1. Change the build project as e.g. above in the "run" Target:

```fsharp
"client", dotnet "fable watch -o output --run webpack-dev-server" clientPath
```

2. Update the entry points in webpack e.g.

```js
fsharpEntry: './src/Client/output/App.js',
```

(Note the lack of `.fs` in the file extension and the extra `output` in the path).

@alfonsogarciacaro is there any reason you would recommend one way or the other with this? What are the pros and cons of both?

isaacabraham:
Fixes #459. This updates the default template to send both src and test outputs into dedicated output folders. It appears to work!


theimowski (approved):
looks good - possibly you want to add the directory to .gitignore?

PR #471, opened 2021-08-20
Remove font awesome from paket

### LINKED ISSUE #456: Consider removing Fable.FontAwesome?
WIth the move to Feliz Bulma, I'm not sure we should still be including the Fable FontAwesome library, unless someone can show how to integrate them together? cc: @Dzoukr

Note: I'm not suggesting moving FA completely, just the F# wrapper.

isaacabraham:
Fixes #456. If there's a way to use these libraries seamlessly with Feliz.Bulma, I'm happy for this not to go in and to add a recipe instead? cc @Dzoukr @Zaid-Ajaj


Dzoukr:
I personally found Font Awesome too quickly evolving to have some strongly-typed API wrapping it so I just use an npm package or direct link to CDN.

isaacabraham:
That's basically what the template does, aside from still include the NuGet package. As we've now moved over to Feliz.Bulma, I'm not sure what sense it makes to keep it.

theimowski:
yeah let's get it removed - noticed there's a recipe in docs that has inconsistent instructions between default and minimal: https://github.com/SAFE-Stack/docs/blob/master/docs/recipes/ui/add-fontawesome.md#3-code-snippet

PR #482, opened 2021-11-22
Include a correctly configured VS launch settings file.

Ryan-Palmer-CIT:
If a launch settings file isn't included, it is created when you first open a new SAFE app in Visual Studio. However, it is created with the wrong ports, and this causes a broken browser tab to be launched when running the app. This often confuses new users and always needs explaining when we run a course. By including a correctly configured file with the template we will ensure users have a better experience out of the box.


isaacabraham:
LGTM

Issue #472, opened 2021-08-20
Update to webpack 5


isaacabraham:

theimowski:
I recall there were same issues for Feliz and Webpack 5 - https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/330 
Not sure if it relates to stuff we have in template though @Zaid-Ajaj

mattgallagher92:
Summary from a discussion between me, @theimowski and @MangelMaxime:

* react-refresh requires that code follows certain conventions, such as not exposing types or functions that are not React components
* Feliz code that generates JavaScript code that doesn't follow those conventions will not be able to benefit from react-refresh
* In that case, the behaviour will be just the same as it was for webpack 4 (tab reload)

In summary, there is no problem moving to webpack 5. [fulma-demo](https://github.com/MangelMaxime/fulma-demo) is already using webpack 5, so we can use that as a template.

theimowski:
Closed by #485

PR #485, opened 2021-11-23
Upgrade webpack and npm deps

Ryan-Palmer-CIT:
This updates all three webpack files (standard, test and minimal) to Webpack 5 format. It updates the npm dependencies to match. I also found that I had to pass the config explicitly to the fable command in the Run target.

**Note** - when testing I found that the tests are currently broken due to #465. I have submitted a fix for this as a separate PR, #484


isaacabraham:
Closes #472

isaacabraham:
cc: @dzoukr would you be able to have a look at this config given your recent work?

Ryan-Palmer-CIT:
Fwiw I took the SAFEr template as my guide so should look pretty familiar. 

I think we might be able to remove the ` hot: true` from the devServer options as I spotted a hint saying it is the default anyway.

The only thing that stumped me for a few minutes was that the webpack config was no longer picked up automatically, but if I passed it in in the same way we do to for the tests then it worked fine.

Ryan-Palmer-CIT:
Also, I did spot `npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!` in the Travis build error, so I guess that the CI machine might need an npm update?

Dzoukr:
> cc: @Dzoukr would you be able to have a look at this config given your recent work?

Looks good to me, but be aware that I am far from being good at this. I rather randomly change things and see if it works 😄

Ryan-Palmer-CIT:
Hi @theimowski, earlier today @isaacabraham asked me to see if you could take a look at the failing pipelines here? As I mention above, perhaps the Travis machine needs an npm update? If there is anything I can do to help just shout :) Cheers!

theimowski:
The Azure Pipelines seem to be failing due to different wording in new version of webpack. The tests look for a specified [phrase in stdout](https://github.com/SAFE-Stack/SAFE-template/blob/master/tests/Tests.fs#L169) which seems to be now `webpack 5.64.2 compiled successfully in 2469 ms` instead of `... : Compiled successfully. ...`
Not sure yet about Travis failures, but could be as you say - due to incompatible node / npm versions - you can try fiddling with [these lines](https://github.com/SAFE-Stack/SAFE-template/blob/master/.travis.yml#L8-L10)

Ryan-Palmer-CIT:
Just updated the stdOut phrase in the tests so hopefully that will sort Devops out. Also just found [this thread](https://github.com/VickScarlet/lifeRestart/issues/176) which suggests it is a node issue as I thought, so I will see if I can get that updated too.

Ryan-Palmer-CIT:
Node version seems update seemed to work, now both pipelines are failing on webpack being passed an unrecognised '-p' option. I'll see if I can fix that too.

Content/default/webpack.config.js
```diff
@@ -80,7 +86,7 @@ module.exports = {
     plugins: isProduction ?
         commonPlugins.concat([
             new MiniCssExtractPlugin({ filename: 'style.[name].[hash].css' }),
-            new CopyWebpackPlugin({ patterns: [{ from: resolve(CONFIG.assetsDir) }]}),
+            new CopyWebpackPlugin({ patterns: [{ from: resolve(CONFIG.assetsDir) }] }),
         ])
         : commonPlugins.concat([
             new webpack.HotModuleReplacementPlugin(),
```
tico321, line 92: Not completely necessary but since we already have `hot: true` in the devServer this plugin will be added by default, and we could remove it (check the tip inside the docs https://webpack.js.org/configuration/dev-server/#devserverhot)

Ryan-Palmer-CIT: Done

Ryan-Palmer-CIT:
Ok @isaacabraham, all checks seem to be passing now

Content/default/webpack.config.js
```diff
@@ -80,24 +86,24 @@ module.exports = {
     plugins: isProduction ?
         commonPlugins.concat([
             new MiniCssExtractPlugin({ filename: 'style.[name].[hash].css' }),
-            new CopyWebpackPlugin({ patterns: [{ from: resolve(CONFIG.assetsDir) }]}),
+            new CopyWebpackPlugin({ patterns: [{ from: resolve(CONFIG.assetsDir) }] }),
         ])
-        : commonPlugins.concat([
-            new webpack.HotModuleReplacementPlugin(),
```
theimowski, line 86: are we removing support for HMR?

Content/minimal/webpack.config.js
```diff
@@ -25,32 +30,39 @@ var copyPlugin =
...
-    publicPath: '/',
-    contentBase: resolve('./src/Client/public'),
+    static: {
+        directory: resolve('./src/Client/public'),
+        publicPath: '/'
+    },
     host: '0.0.0.0',
     port: 8080,
-    inline: true,
     proxy: {
```
theimowski, line 39: I'm getting following error when testing minimal template:

```
<e> [webpack-dev-server] [HPM] Error occurred while proxying request localhost:8080/api/hello to http://localhost:8085/ [ECONNREFUSED] (https://nodejs.org/api/errors.html#errors_common_system_errors)
```

maybe dev server proxy configuration changed format with new WebPack version?

Content/default/webpack.config.js
```diff
@@ -24,21 +29,17 @@ var CONFIG = {
         // redirect requests that start with /api/ to the server on port 8085
```
theimowski, line 29: same as with minimal, dev server proxy doesn't seem to work with default template:

```
client: <e> [webpack-dev-server] [HPM] Error occurred while proxying request localhost:8080/api/ITodosApi/getTodos to http://localhost:8085/ [ECONNREFUSED] (https://nodejs.org/api/errors.html#errors_common_system_errors)
```

and before that client prints following to stderr:

```
client: <i> [webpack-dev-server] [HPM] Proxy created: /api/**  -> http://localhost:8085
client: <i> [webpack-dev-server] [HPM] Proxy created: /socket/**  -> http://localhost:8085
client: <i> [webpack-dev-server] Project is running at:
client: <i> [webpack-dev-server] Loopback: http://localhost:8080/
client: <i> [webpack-dev-server] On Your Network (IPv4): http://10.9.136.121:8080/
client: <i> [webpack-dev-server] On Your Network (IPv6): http://[fe80::1]:8080/
client: <i> [webpack-dev-server] Content not from webpack is served from '/Users/theimowski/oss/SAFE-template/Content/default/src/Client/public' directory
client: <i> [webpack-dev-server] 404s will fallback to '/index.html'
```

theimowski (changes requested):
Seems dev server proxy stopped working - see comments

Content/default/webpack.config.js
tico321: Now it's added by default by webpack when you add `hot: true` to the devServer, check the docs  https://webpack.js.org/configuration/dev-server/#devserverhot there is a tip that details this.

Content/default/webpack.config.js
Ryan-Palmer-CIT: That's odd, it seemed to work for me yesterday when I tried it out before pushing. I'll investigate, thanks.

Ryan-Palmer-CIT: @theimowski do you have the latest LTS release of node installed? Just trying to work out why you are seeing these issues as I'm not. Are you running the projects from the Content folder directly, or packaging first or perhaps installing the template?

theimowski: I'm running node 17.0.1 and npm 8.1.0 on MacOS, and testing directly from Content directory

theimowski: Right, I have now downgraded to node 16.13.1, and all works fine. I think I upgraded node recently for some reason to v17, but as v16 is the LTS we could force it for SAFE template - see #489

theimowski (approved):
The issue I referred to before was related to node v17

Content/default/webpack.config.js
Ryan-Palmer-CIT: Ah excellent, glad we worked it out. Maybe it is worth a note somewhere (quickstart?) that warns people of this as I am sure many will hit it.

olivercoad: Can confirm, it works with node `16.13.1` but not `17.2.0`. The client prints the same lines (`<i>`) to stderr in both cases.

olivercoad: Looks like it's probably due to this https://github.com/nodejs/node/issues/40537

The problem is that when the dev server proxies to `http://localhost:8085`, in node 16 it would use ipv4 but in node 17 it defaults to ipv6. However, the server is listening only on ipv4 (`0.0.0.0`).

The fix is to make the server listen on all interfaces, including ipv6. So change `url` in `Server.fs` to `url "http://*:8085"`

Ryan-Palmer-CIT: If one of you could try that and tell me if it works, I'll make the change and push it up. Thanks!

theimowski: see https://github.com/SAFE-Stack/SAFE-template/pull/490

Issue #462, opened 2021-08-09
Upgrade to Chockidar 3


isaacabraham:
Building a SAFE v3 project currently gives the following:

```
chokidar@2.1.8: Chokidar 2 will break on node v14+. Upgrade to chokidar 3 with 15x less dependencies.
```

I got stung by this recently. People that have newer versions of node will simply get a crappy error message on first time they use SAFE.

theimowski:
I still can't get my head around npm dependencies - we actually have chokidar pinned to version 3.5.1. I assume that this warning comes from one of the transient dependencies - e.g. `webpack` -> `watchpack` -> `watchpack-chokidar2` -> `chokidar ^2.1.8` 
Not sure if one can mute such warnings given that we already use newer version, or whether the only solution is to update webpack family to version 5

isaacabraham:
Was Chokidar always pinned? Let me test this again with newer Node - maybe I had an older version of the template running.

isaacabraham:
Fixed by #472

PR #486, opened 2021-11-25
Temp fix - chokidar

Akash-Mair:
On the minimal template if you have a specific version of node(14.17.2/npm(7.19.1) you get the below error

```bash
throw new Error(
^

Error: No version of chokidar is available. Tried chokidar@2 and chokidar@3.
You could try to manually install any chokidar version.
chokidar@3: Error: Cannot find module 'chokidar'
```

But the full template is fine, this is because in the full template various dependencies (sass) use the latest version of chokidar and webpack is using 4.44.6

I've copied over the same versions of webpack/webpack-cli/webpack-dev-server from the full to the minimal as they should always be in sync and re made the lock file. 

This has fixed the bug :)

### DIFF
--- a/Content/minimal/package-lock.json
+++ b/Content/minimal/package-lock.json
@@ -5,20 +5,26 @@
     "packages": {
         "": {
             "devDependencies": {
-                "copy-webpack-plugin": "^6.4.0",
+                "copy-webpack-plugin": "^6.1.0",
                 "html-webpack-plugin": "^3.2.0",
-                "webpack": "^4.44.2",
-                "webpack-cli": "^3.3.12",
-                "webpack-dev-server": "^3.11.0"
+                "webpack": "^4.41.2",
+                "webpack-cli": "^3.3.9",
+                "webpack-dev-server": "^3.8.2"
             }
         },
+        "node_modules/@gar/promisify": {
+            "version": "1.1.2",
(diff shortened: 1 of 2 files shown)


Akash-Mair:
@isaacabraham

isaacabraham:
Fixes #462

isaacabraham:
@theimowski you ok with this?

theimowski:
Is this maybe fixed by #485?

Ryan-Palmer-CIT:
This is indeed fixed by the webpack update. I think the plan was to fast track in this fix whilst we work on getting #485 merged

theimowski:
yeah I'd say let's get #485 to work, don't think we need to rush with this workaround

theimowski:
can we close it now?

halcwb:
The above problem is not resolved in 3.1.1 is this correct? Got it working upgrading webpack and copying dependencies from the latest package.json from the full safe-stack template.

isaacabraham:
@halcwb Yes, that looks to be the case. @theimowski what's your thoughts on release e.g. SAFE 3.2 (or 3.5?) which upgrades this?

isaacabraham:
I think we can close this now, especially since #485 has been merged in.

theimowski:
yeah we should release new version soon (quite a few changes) but I think it makes sense to include .net 60 upgrade https://github.com/SAFE-Stack/SAFE-template/pull/494

isaacabraham:
Fixed by #472

PR #490, opened 2021-12-07
Listen on all interfaces including ipv6

olivercoad:
Fixes localhost proxy for node 17, discussion https://github.com/SAFE-Stack/SAFE-template/pull/485#discussion_r760543025


theimowski:
thanks!

PR #491, opened 2021-12-08
Webpack improvements

### LINKED ISSUE #488: Updates to webpack.config.js and webpack.tests.config.js (discussion)
Hey guys, recently I recreated the template from scratch https://github.com/tico321/SAFEFromScratch

And there are a few changes I would like to bring to the webpack config files so they are more aligned with js best practices, but before I create a PR please let me know your thoughts about the following changes:

- Use const instead of var to declare variables.
- Update the webpack.tests.config.js file to have only what's necessary for testing as there is no need to have production building logic in there.
- Add some tasks that we can run with  `npm run` and invoke them from Fake (see this [package.json](https://github.com/tico321/SAFEFromScratch/blob/master/package.json) for an example), so instead of running:
```
"client", dotnet "fable watch -o output -s --run webpack-dev-server --config ../../webpack.tests.config.js" clientTestsPath
```
we would run:
```
"client", dotnet "fable watch -o output -s --run npm start:tests --prefix ../.." clientTestsPath
```
- Identify the environment based on the value coming from `process.env.NODE_ENV` if you reviewed the package.json above this is done by passing the `--node-env` flag.
- Move the bulma, react and react-dom to dependencies instead of devDependencies in the package.json file

tico321:
I want to propose the following changes to webpack in order to align with common practices in the JS world. Here is the list of changes:
- Use `const` instead of `var` when possible.
- Split `devDependencies` and `dependencies` in the package.json file.
- Use script commands from Fake instead of calling webpack directly.
- Make config and test config share the same logic to avoid having to update both configurations every time. 
- Identify the environment and mode through environment variables and arguments passed to webpack.

Some additional changes:
- I updated the copy-webpack-plugin to the latest version (10.0.0)
- I updated `[hash]` to `[contenthash]` for the mini-css-extract-plugin and for the js output since `[hash]` has been deprecated.

I'll be happy to hear your feedback on this proposal, so please let me know if anything can be improved. 

closes #488


Content/default/webpack.config.js
```diff
@@ -28,113 +17,37 @@ var CONFIG = {
...
-    new HtmlWebpackPlugin({
-        filename: 'index.html',
-        template: resolve(CONFIG.indexHtmlTemplate)
-    })
-];
+module.exports = function(env, arg) {
+    // Mode is passed as a flag to npm run. see the docs for more details on flags https://webpack.js.org/api/cli/#flags
+    const mode = arg.mode ?? 'development';
+    // environment variables docs: https://webpack.js.org/api/cli/#environment-options
+    const config = env.test ? TEST_CONFIG : CONFIG;
```
theimowski, line 48: If you have those two configurations in same file, why the extra `webpack.common.js` file? Would be simpler to just have everything inlined into single file right?

tico321: Yeah, I was thinking about it as well. I was not sure which one would look cleaner. 
Also ehere would you place the build function, I was thinking two options:
Option 1:
```javascript
const CONFIG = { /*...*/}
const TEST_CONFIG = {/*...*/}
module.exports = function(env, arg) { /*...*/ }

//build function at the bottom
function buildConfig(config, mode) { /*...*/ }
```

Option 2:
```javascript
const CONFIG = { /*...*/}
const TEST_CONFIG = {/*...*/}
function buildConfig(config, mode) { /*...*/ }
// module.exports as last statement
module.exports = function(env, arg) { /*...*/ }
```

any preference?

tico321: I pushed a commit with Option 1. Please let me know if you have any feedback.

Content/default/webpack.config.js
```diff
@@ -49,5 +47,112 @@ module.exports = function(env, arg) {
...
+        //      - HotModuleReplacementPlugin: Enables hot reloading when code changes without refreshing
+        plugins: isProduction ?
+            commonPlugins.concat([
+                new MiniCssExtractPlugin({ filename: 'style.[name].[contenthash].css' }),
+                new CopyWebpackPlugin({ patterns: [{ from: resolve(config.assetsDir) }] }),
+            ])
+            : commonPlugins,
+        resolve: {
+            // See https://github.com/fable-compiler/Fable/issues/1490
+            symlinks: false
```
olivercoad, line 111: Does anyone know if this is still needed when not using `fable-loader` webpack loader?

Content/default/webpack.config.js
olivercoad: If you're asking, I think option 2. Or just not have a `buildConfig` function at all and just do it all inline.

olivercoad:
Has anyone tried writing their Webpack config in F# yet? 🤪

Content/default/webpack.config.js
tico321: I just pushed a commit following your suggestion.
To be honest I don't have a particular preference, all these options look good to me. But I'm happy to try alternatives until we find one that looks good for all of us.

tico321:
> Has anyone tried writing their Webpack config in F# yet? 🤪

Not yet, but could be interesting. I'm thinking that using anonymous records could be the easiest thing to do.

theimowski:
Looks good :+1: @Zaid-Ajaj @MangelMaxime can you have a look please?

MangelMaxime:
All the stuff listed in @tico321 comments seems goods to me.

One tips, that I like to use now days for the plugins array is to use `.Filter(Boolean)` as I find it easier to read and add new plugins depending on the target:

```js
        plugins:
            [
                new HtmlWebpackPlugin({
                    filename: "./index.html",
                    template: "./src/index.html"
                }),
                new MiniCssExtractPlugin(),
                isProduction && myProductionOnlyPlugin() // This plugin will only be added for production
            ].filter(Boolean),
```

But it depends on what people find here to read between that and the concat. 

About the `symlink` comment, I suppose it is not required anymore has Fable doesn't retrieve the file via webpack anymore.

olivercoad:
Yep, I think I like that. That way you don't have to scroll up to find `commonPlugins` half a screen earlier. Not quite as clean as F# (implicit) yields though 😅. It does use some javascript idiosyncrasies that not everyone might be familiar with, but I think with a comment there it's probably fine.

Content/default/webpack.config.js
```diff
@@ -28,113 +15,139 @@ var CONFIG = {
...
-        ]
-    }
+        },
+        // Besides the HtmlPlugin, we use the following plugins:
+        // PRODUCTION
+        //      - MiniCssExtractPlugin: Extracts CSS from bundle to a different file
+        //          To minify CSS, see https://github.com/webpack-contrib/mini-css-extract-plugin#minimizing-for-production
+        //      - CopyWebpackPlugin: Copies static assets to output directory
+        // DEVELOPMENT
+        //      - HotModuleReplacementPlugin: Enables hot reloading when code changes without refreshing
```
olivercoad, line 97: This comment is out of date, `HotModuleReplacementPlugin` is not used.

tico321:
As it looks like we no longer need the `symlink` I just removed it.
I also played a little bit with the filter option that @MangelMaxime suggested and I think it does look cleaner, so I added that modification as well (Note that I also updated the comments and placed them above the plugins). 
Any feedback is welcome, so please let me know if it looks awkward and I can rollback to use a common plugins variable.

tico321:
Is anything preventing the PR from being merged? Please let me know if I missed something

theimowski:
Sorry, didn't have time to get to this up till now.
This looks awesome - many thanks! 
There are a few things we still need to wrap up before releasing new template version

PR #495, opened 2021-12-14
Fix Client.Test.fs Expect.equal parameters order

linitachi:
In line 15, actual value is model.Todos.Length , expected value is 1.
In line 16, actual value is model.Todos.[0] , expected value is newTodo.
Both origin code  are reverse parameters.


theimowski:
They were indeed, thanks!

Issue #477, opened 2021-09-29
Move default deployment to Linux OS


Akash-Mair:
Linux based apps are cheaper to run. We can create a recipe on how to move to Windows OS.

isaacabraham:
I'm fully in favour of this - Linux is around 4x cheaper than Windows on App Service.

This would need update to Farmer and some minor changes to the template e.g. port number but otherwise not a lot. @theimowski what do you think?

The only other things we need to change are:

* Logging: We can't use the app service logging extension any more
* App Insights: We'll need to explicitly hook into AI in the server rather than the code-free approach now. This is a one-liner but needs doing.

theimowski:
Fine with me 👍

isaacabraham:
Other observation we've learned recently:

1. Only one free Linux service plan per region / subscription. Unlimited free on Windows.

theprash:
I have a new PR #528 (which I propose we use instead of #498) which makes the changes described in my previous comment but doesn't actually set Linux as the OS.

I'll open a PR to update the docs. We might need to mention in the docs that there are 2 possible ports depending on your SAFE template version.

theprash:
Do we definitely want to enable Linux by default given the restriction to just one free Linux instance? The good news is that after my PR, the recipe to switch to Linux is a 1-line change.

Another option is we stick to Windows as a default but add it explicitly in Farmer rather than relying on an implicit default. That way it might more obvious how to switch to Linux and we wouldn't even need a recipe.

PR #498, opened 2022-01-14
Linux deployment 

Akash-Mair:
This PR solves issue #477 

I've updated the default template:

Files changed:
- `Server.fs` and `webpack.config.js` : Port number
- `Build.fs` : Farmer script

I've tested locally as well as deploying to Azure 

@isaacabraham  let me know if I've missed anything or if anything needs to be changed :)


isaacabraham:
We need to wait on a Farmer update first.

theprash:
I propose that we abandon this in favour of #528

Issue #460, opened 2021-08-04
Dotnet run opens the wrong URL


inputusernamehere:
After creating a new project with the template, running the command `dotnet run` results in my browser opening a tab with the url `http://0.0.0.0:8085/`. However this is the backend's URL. Shouldn't it open the client instead, at `:8080`?

theprash:
The template stopped automatically opening a browser tab a while back because it's annoying while working on a project when you already have the browser tab ready to go.

Do you use Visual Studio? It creates `launchSettings.json` with `"launchBrowser": true`. This then causes dotnet to launch the browser on run. I'm not sure how to correct the port here but you can at least change `launchBrowser` to false.

inputusernamehere:
Ahh must be the case yeah, I do indeed have a launchSettings.json file with `"launchBrowser": true`. Case solved. Maybe it would be better for the template to include a launcSettings.json where that setting is set to false?

isaacabraham:
@theprash I've seen this too. Either including a file that stops this happening, or including documentation to explain this, would be my suggestion.

theprash:
I found a better fix for this issue than including a default file. You can add `<NoDefaultLaunchSettingsFile>true</NoDefaultLaunchSettingsFile` to Server.fsproj. I opened PR #523 to make this change to both default and minimal templates.

theprash:
Actually cancel that (my last comment). If we remove the file then the environment is defaulted to Production while running in dev. So I've opened PR #524 to add launchSettings.json to the minimal template as well.

PR #523, opened 2022-09-05
Use <NoDefaultLaunchSettingsFile/> instead of creating a default file

theprash:
A better fix for #460 applied to both default and minimal templates.


theprash:
Actually, I'm going to close this. By removing the launchSetttings.json file the server defaults to running in "Production" environment during dev.

PR #526, opened 2022-09-05
Add .paket/ to .gitignore

mattgallagher92:
As recommended at https://fsprojects.github.io/Paket/get-started.html.

> Make sure to add the following entries to your `.gitignore`:
>  ```gitignore
>  #Paket dependency manager
>  .paket/
>  paket-files/
>  ```

### DIFF
--- a/Content/default/.gitignore
+++ b/Content/default/.gitignore
@@ -11,6 +11,7 @@ obj/
 bin/
 output/
 packages/
+.paket/
 paket-files/
 node_modules/
 release.cmd


olivercoad:
Not sure that documentation is correct. I was always under the impression that `.paket/Paket.Restore.targets` is meant to be committed for VS or other msbuild things that don't know about paket. See https://github.com/fsprojects/Paket/issues/2840

theprash:
I tried deleting `.paket/Paket.Restore.targets` and it breaks `dotnet run`, so clearly it's important and should be version controlled. It's also explicilty included in the template, which supports this further.

So we probably don't want to add `.paket/` to gitignore so that the file gets included on first commit.

I believe in older versions of Paket, there used to be at least one other file that Paket would create in the directory, which should be git ignored. So in that situation you should probably ignore `.paket/` but then add an exception to un-ignore `.paket/Paket.Restore.targets`. However, I've not seen that extra file around for quite a while so I assume it's no longer created by Paket, and the Paket site is simply out of date.

Was there a problem that motivated this change?

mattgallagher92:
Thanks both. It seems that this change would be harmful then. I'll close the PR without merging.

@theprash the motivation came from someone asking me whether they should commit `.paket/Paket.Restore.targets` and I assumed that I could trust the Paket documentation.

PR #528, opened 2022-09-06
Changes to allow Linux Azure App Service

theprash:
Closes #477

PR #530, opened 2022-09-16
Fix dev-time server URL (https -> http)

theprash:
Fixes the issue reported by @theimowski in #528


theimowski:
this works ok locally, just wondering if it has any implications on the Azure side of things? Don't really have a chance to test anymore Azure deployment. 

I was wondering why the tests didn't catch the issue, and realised they don't really test HTTP requests against server, so as a quick fix I came up with this: https://github.com/SAFE-Stack/SAFE-template/pull/531

theprash:
@theimowski launchSettings.json is for dev time only. I've just checked anyway and the Azure deploy still works on Windows and Linux.

Issue #533, opened 2022-10-01
Having trouble running the new template


hede:
Hello,
I just installed the 4.1.0 template.
dotnet run does not work out of the box.

The client fails with a bunch of "namespace or module not defined" errors:
```
.\Index.fs(3,6): (3,12) error FSHARP: The namespace or module 'Elmish' is not defined. (code 39)
.\Index.fs(4,6): (4,11) error FSHARP: The namespace or module 'Fable' is not defined. (code 39)
...
```

The server seems to run. It stands a web server up at localhost:5000, but when I go there it's 404 (last 4 lines of the log).

Please let me know if there is something else you want me to try or to see the output of. 

## full log:
```
D:\src\SafeTest>dotnet new SAFE
The template "SAFE-Stack Web App v4.1.0" was created successfully.


D:\src\SafeTest>dotnet tool restore
Tool 'paket' (version '7.0.2') was restored. Available commands: paket
Tool 'fable' (version '3.7.5') was restored. Available commands: fable
Tool 'fantomas-tool' (version '4.7.1') was restored. Available commands: fantomas
Tool 'femto' (version '0.12.0') was restored. Available commands: femto

Restore was successful.

D:\src\SafeTest>dotnet run
run Run
Building project with version: LocalBuild
Shortened DependencyGraph for Target Run:
<== Run
   <== InstallClient
      <== Clean

The running order is:
Group - 1
  - Clean
Group - 2
  - InstallClient
Group - 3
  - Run
Starting target 'Clean'
D:\src\SafeTest\src\Client> "dotnet" fable clean --yes (In: false, Out: false, Err: false)
Fable: F# to JS compiler 3.7.5
Stop Russian aggression against Ukraine!

This will recursively delete all *.fs.js[.map] files in D:\src\SafeTest\src\Client
No files have been deleted. If Fable output is in another directory, pass it as argument.
Finished (Success) 'Clean' in 00:00:00.7408609
Starting target 'InstallClient'
.> "C:\Program Files\nodejs\npm.CMD" install (In: false, Out: false, Err: false)
npm WARN config global `--global`, `--local` are deprecated. Use `--location=global` instead.
npm WARN deprecated querystring@0.2.1: The querystring API is considered Legacy. new code should use the URLSearchParams API instead.
[...]

PR #534, opened 2022-10-05
Fixes .net sdk selection issue discussed in #533

hede:
It looks like the sdk version needs a band/feature version to roll forward *from*


theimowski:
Thank you!

PR #541, opened 2022-11-18
Update package.json files to allow node 18 (as well as 16)

theprash:
The templates work with node v18 (current latest LTS), so we could allow it.


theprash:
@theimowski Could we get a release for this? If it could be today that would be very useful! Thank you 😄

theimowski:
yes, released now in 4.2.0, thanks!

PR #542, opened 2022-12-01
Update package.json to allow node 19

johanvts:
To support node 19 (and 16 and 18).

### DIFF
--- a/Content/default/package.json
+++ b/Content/default/package.json
@@ -1,7 +1,7 @@
 {
     "private": true,
     "engines": {
-        "node": "~16 || ~18",
+        "node": "~16 || ~18 || ~19",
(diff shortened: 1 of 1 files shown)


theprash:
Thanks @johanvts. I tested this by installing node 19 and running the template and I got this error:

```
npm ERR! notsup Not compatible with your version of node/npm: undefined
npm ERR! notsup Required: {"node":"~16 || ~18 || ~19","npm":"~8"}
npm ERR! notsup Actual:   {"npm":"9.2.0"}
```

Looks like we should also allow the npm 9 because it's installed by default with node 19.

isaacabraham:
Will close in a week - can review upgrading as part of v5.

Issue #529, opened 2022-09-12
Move from fantomas-tool to fantomas


isaacabraham:
With the imminent release of Fantomas 5, we should upgrade the `dotnet-tools.json`:

```diff
-"fantomas-tool": {
+"fantomas": {
-  "version": "4.7.1",
+  "version": "5.0.0",
  "commands": [
    "fantomas"
  ]
},
```

We also need to update `.editorconfig` as follows:

```diff
-[src/Client/*.fs]
-# Feliz style
-fsharp_single_argument_web_mode=true
+fsharp_multiline_block_brackets_on_same_column = true
+fsharp_experimental_stroustrup_style = true
```

isaacabraham:
@nojaf is this roughly correct?

nojaf:
Yes, that will do the trick.

On a side note:
I would also advise to change `insert_final_newline = false` to `true`.
See [link](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline)

isaacabraham:
Will do. Any reason why this isn't the default then? :-)

nojaf:
It is the default in Fantomas, but the `.editorconfig` of the template seems to have put it on `false`.

PR #543, opened 2022-12-05
529 move from fantomas-tool to fantomas

OutOfScopeia:
Issue #529


theprash:
I'm closing this to replace it with #548 which also contains these commits from a different fork.

PR #549, opened 2023-02-16
micro event sourcing support

tonyx:
Added a minimal set of libraries supporting event sourcing. Added the option of managing the addTodo/removeTodo as events stored in specific postgres db or in memory.


theprash:
Thanks for contributing this @tonyx. It's not going to be part of the main SAFE template but the change you've made could be a useful sample for someone interested in using event sourcing with SAFE.

Keeping it in PR form is not particularly useful since the folder structure was changed and so all the code appears as new files. So I would suggest that you keep it as a sample project in your own forked version of the repo, and possibly rename the repo accordingly. I will close this PR for now.

Once it's ready, maybe there would be somewhere good to mention your repo in the [SAFE stack docs](https://safe-stack.github.io/docs).

Issue #487, opened 2021-11-25
Specify major version in paket.dependencies


olivercoad:
So that things don't break when dependencies make breaking changes and bump the major version number.

Using `~>` operator, as discussed in #180

theimowski:
I'm not convinced we should do this - yes on one hand you might get breaking changes, but on the other hand if you make `~>` a default, one would need to manually remove it to get a new major version.

theprash:
One option is to use the [`--keep-major`](https://fsprojects.github.io/Paket/paket-update.html) option on Paket's update, add and install commands. It makes Paket "only allow updates that preserve the major version".

theprash:
We have a workaround and no clear path forward here so I'll close the issue.

olivercoad:
Thanks all for the discussion on this over in #553 and thanks @Tom-Sloboda for making the PR (#559). I agree that it's a good thing that you'd have to manually change it to get a new major version, because if a library is explicitly saying it has a breaking change by incrementing the major version, then I usually want to know about it and take the time to look at the changelog to find out what (if any) other changes I need to make while updating it.

mattgallagher92:
Merged PR.

PR #559, opened 2023-04-21
#487 Specify major version in paket.dependencies

Tom-Sloboda:


mattgallagher92:
Looks good to me. Would appreciate explicit go ahead from @isaacabraham and @theimowski.

mattgallagher92:
Merging as there haven't been any objections.

PR #566, opened 2023-09-29
Fix issues in default template build script for deploy to Azure

martinbryant:


martinbryant:
Hi @theimowski it would be great if at some point a release of the template could be organised and I am happy to have a conversation with you about how this works so that I can organise it in the future.

Content/default/Build.fs
```diff
@@ -23,14 +23,14 @@ Target.create "Clean" (fun _ ->
 Target.create "InstallClient" (fun _ -> run npm [ "install" ] ".")
 
 Target.create "Bundle" (fun _ ->
-    [ "server", dotnet [ "publish"; "-c"; "Release"; "-o"; $"\"{deployPath}\"" ] serverPath
+    [ "server", dotnet [ "publish"; "-c"; "Release"; "-o"; deployPath ] serverPath
```
theprash, line 26: Please could you confirm or verify that this still works if the deploy path has spaces in it? Thanks.

martinbryant: Screenshot 2023-09-29 at 10 44 45

Yes this has been checked as per the above - the path for the command has been wrapped in quotes allowing for spaces

theprash (approved):

PR #568, opened 2023-10-06
roll default template to net8

martinbryant:


martinbryant:
Deleted as was only used to test the deployment away from upgrading the template

PR #570, opened 2023-11-24
Update fantomas config and format files

martinbryant:
- Update fantomas config to format brackets
- Format the files
- Change className to classes for styles so formatting is nicer


mattgallagher92 (approved):
LGTM 👍

PR #571, opened 2023-11-28
add fantomas CE no new line to config

martinbryant:


mattgallagher92 (changes requested):
Think you need to run this too right? E.g. Build.fs has a newline before the expression.

mattgallagher92 (approved):
LGTM 👍

PR #574, opened 2023-11-29
removed -r switch from fantomas/run at root in build.fs

martinbryant:
Only just noticed that the -r created a folder in src as I ran dotnet run format yesterday ;) 

Should run on Build.fs also


isaacabraham:
I think personally having this target is overkill but yeah it should now work :-)

Issue #552, opened 2023-03-05
Consider moving build sln to a build folder


halcwb:
Currently, the build sln sits right at the root folder of the template. Wouldn't it be better to move this to a 'build' folder instead? Only, then you can't just `dotnet run`, but fixing this in the package.json and using `npm start` would be also be convenient.

mattgallagher92:
Thanks for the suggestion @halcwb! I agree that it's a little untidy to have the build project's files right at the repo root.

My opinion is that `dotnet run` is more natural than `npm start` for a few reasons:

- The build project is actually defined in .NET.
- The code is compiled using dotnet tools, hence dotnet feels more integral to the stack.
- `npm start` would mean one more layer of indirection, which might be confusing to newcomers. There's already a lot to take in if you're only familiar with one of the .NET or JavaScript ecosystems.
- Ultimately, the application will be running in a .NET environment.

However, I don't feel particularly strongly about any of those points.

A compromise approach could be to leave Build.fsproj where it is, but move Build.fs and Helpers.fs into a subdirectory (for example `./build/` or `./src/build/`). What do you think?

halcwb:
@mattgallagher92 
For a template recognizability is indeed very important. To just create a starting point using a `build.cmd` or `build.sh` would also work I think. But this is more a subjective taste issue? 

With the current setup the problem I have is when I do a `dotnet build` or `dotnet test` that's not working out of the box because then I have to specify the `sln` I want to use.

mattgallagher92:
At Compositional IT, we'd usually approach that by having dedicated build targets in the build project. There is already a `"RunTests"` build target in the template, executable using `dotnet run RunTests`: https://github.com/SAFE-Stack/SAFE-template/blob/91c167293cddc63f474caf911907fefa245f3bba/Content/default/Build.fs#L55-L60

You could create your own targets, for example `"BuildSln"` and `"TestSln"` that perform `dotnet build` and `dotnet test` respectively against the solution.

That way the build project always behaves as the target of your `dotnet run ...` commands, and is always the command-line entry point into actions against your codebase. Build.fs then captures the actual behaviour of those commands somewhat analogously to how build.cmd or build.sh would.

mattgallagher92:
@isaacabraham @theprash I'd be interested in your thoughts regarding this:
> leave Build.fsproj where it is, but move Build.fs and Helpers.fs into a subdirectory

theprash:
[...]

PR #576, opened 2023-11-29
Replace FAKE with Fun.Build.

isaacabraham:
This (somewhat experimental) PR replaces FAKE with [Fun.Build](https://github.com/slaveOftime/Fun.Build/) for the build project. The benefits of this PR are:

* Approximately 20% reduction in lock file (185 -> 150 LOC) - smaller lock file means quicker resolutions etc.
* Approximately 40% reduction in build project code (185 -> 105 LOC)
   * Part of this effort also includes simply removing some unnecessary symbols that were only used in one place, as well as simplifying each pipeline (think: FAKE target) to be more self contained.
   * Code is simpler, with no need for a Helpers file.
   * Code is easier to read and reason about.

I also have used this opportunity to only run a "clean" on doing a Bundle - there's no need to do it on a Run as far as I can see.

You can still run the "default" pipeline with a standard `dotnet run`. Running other pipelines uses `dotnet run -- -p <pipeline name>`.


martinbryant:
- Like it for the reasons you have mentioned
- Are we considering it for the initial release or should we do it after?
- Is the running of other targets being different syntax an issue and/or can we consider looking at PR to library to make it work as it is currently?
- Work to be done to sort the template test by the looks of it
- What no colours?? ;) Again a PR or something to add this as with multiple targets I client/server/docker I find this really useful?

mattgallagher92:
1. The code is much nicer (apart from the hideous "paralle" (sic.) keyword) - what happens for each target is so much clearer.

1. > Is the running of other targets being different syntax an issue and/or can we consider looking at PR to library to make it work as it is currently?

   @martinbryant you mean compared to FAKE? It's not a problem so long as it's clearly documented. The code comments  make that clear, but updating the README would help too.

1. For me the colours aren't a dealbreaker, and I'd consider releasing the template without colours but, as you say, a PR to Fun.Build would make sense.

mattgallagher92:
Reasons not to include this in SAFE v5.0.0 (but consider for v5.1.0):

* We know that FAKE extends well to other tools like Docker. I'd want to test drive Fun.Build a bit before including it in the template
* For feature parity, we probably should get console colours working

Reasons to prefer FAKE anyway:

* FAKE is well established and battle-tested
* FAKE has thorough docs
* FAKE has a large pool of contributors, so I have more confidence that it will be actively maintained for a long time yet

isaacabraham:
@martinbryant I've fixed the test build now, this is now green.

martinbryant:
Think we can look at closing #552 when completing this

isaacabraham:
Is this generally accepted as "the way forward"? If so, I'll resolve the conflicts and get this ready on Friday.

mattgallagher92:
I'm ambivalent. I don't feel any strong need to move away from FAKE, but the benefits you listed seem good. Moving back in future should be relatively easy if necessary.

Given that we're changing how programs like dotnet and npm are called, we should test on Windows, Mac and Linux before merging.

mattgallagher92:
Resurrecting this discussion @isaacabraham.

I'm okay to switch, but I do think the lack of colouration might be pain for some users (us included). How nice would the code be if we wanted to share stages across multiple pipelines? E.g. `docker compose up` is a prerequisite for multiple operations we run in some codebases.

@jwthomson @Larocceau do you have any strong feelings?

PR #577, opened 2023-12-01
used className attribute for styles in Index.fs

martinbryant:


Content/default/src/Client/Index.fs
```diff
@@ -92,18 +68,10 @@ let private todoAction model dispatch =
...
-            "shadow-md"
-            "p-4"
-            "w-5/6"
-            "lg:w-3/4"
-            "lg:max-w-2xl"
-        ]
+        prop.className "bg-white/80 rounded-md shadow-md p-4 w-5/6 lg:w-3/4 lg:max-w-2xl"
         prop.children [
             Html.ol [
-                prop.className [ "list-decimal"; "ml-6" ]
```
jwthomson, line 106: Not sure how this was working before - does className have an overload to accept a list?

jwthomson: Not that it matters now! 😄

theprash: Must've been that, yes. Feliz uses methods with lots of overloads. Very handy.

Issue #580, opened 2023-12-01
Look at files in Client.fsproj that probably are not required to be


martinbryant:
The following files are noted in the Client.fsproj that are dev/build specific and do not affect the building of the Client project.

- vite.config.mts 
- tailwind.config.js 
- postcss.config.js

We should look to get input from @MangelMaxime about this and if we can remove them if there is no reason to leave them

MangelMaxime:
Hello,

Fable only needs the `.fs` files to be include in the project file.

Adding others files likes:

```xml
        <None Include="postcss.config.js" />
        <None Include="tailwind.config.js" />
        <None Include="index.html" />
        <None Include="paket.references" />
```

Is a matter of preference.

Personally, because I never use the Solution Explorer view, I never add this files in my project files. But I know people who use Solution Explorer instead of File Explorer, includes this files so they can have an easy access to them.

The decision on this subject, depends on how you think people will consume the template and use it.

theprash:
These are there because I am intentionally including them to make the files easily available in Visual Studio. Is there some reason to remove them?

RicoSaupe:
Updated the proj and sln files in my pr. Please have a look and approve.

PR #582, opened 2023-12-02
updated sln reference, readme

RicoSaupe:
first small PR. :) 

in regards to issue #580 i discovered also the following. 

1. webpack config was still in default sln file
2. travis.yml still referenced net 6
3. updated readme.md files to reflect used net8 and updated npm, node versions


RicoSaupe:
Another thought. Do we need to reference these file as solution files? They are generated anyways and not to be edited or is someone changing them? 

package-lock.json
paket.lock

isaacabraham:
@martinbryant these changes look good, especially the `readme.md` which is critical to get correct - it's the docs file that everyone looks at the first time.

Content/default/README.md
```diff
@@ -17,6 +18,15 @@ Before you run the project **for the first time only** you must install dotnet "
 dotnet tool restore
 ```
 
+To check for updates, use
```
isaacabraham, line 21: I'm not sure that this additional section is required - it's just standard `dotnet tool` information.

Content/minimal/README.md
```diff
@@ -26,6 +27,15 @@ dotnet tool restore
 dotnet fable watch src/Client --run npm run start
 ```
 
+To check for updates, use
```
isaacabraham, line 30: Ditto above comment.

.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: csharp
-dotnet: 6.0.101
+dotnet: 8.0.100
```
isaacabraham, line 2: Definitely want this.

Content/minimal/.gitignore
```diff
@@ -13,5 +13,6 @@ node_modules/
 release.cmd
 release.sh
 *.orig
+*.fsproj.user
 *.DotSettings.user
-*.fs.js
\ No newline at end of file
+*.fs.js
```
isaacabraham, line 18: We probably want `fs.js` in both minimal and full.

Content/default/SAFE.App.sln
```diff
@@ -9,7 +9,6 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 		paket.lock = paket.lock
 		package-lock.json = package-lock.json
 		package.json = package.json
-		webpack.config.js = webpack.config.js
```
isaacabraham, line 12: Yes, remove this - not required (in fact, it's wrong to have it there)..

isaacabraham (changes requested):
I think that this is a good set of changes, especially the readme.

Content/default/README.md
RicoSaupe: correct, is just what you should know anyways. I felt it was kind of helpful for me lately. But you are right.

Content/minimal/.gitignore
RicoSaupe: agree, not sure why this was removed actually

Content/default/README.md
RicoSaupe: removed this section

Content/minimal/README.md
RicoSaupe: removed this section

RicoSaupe:
@isaacabraham was aligning the project file names to start upper case but if failed here "tests/tests.fsproj". I could rename the file and also the references to it but git ignored the upper lower case file name and did not push the file name change. So you build failed, since its using ubuntu. 
Any idea how to push also the file name changes? If you think its ok.

martinbryant:
@RicoSaupe thanks vert much for the PR. 

I have pushed some changes that fixed the file renaming. Its a bit silly that it doesn't recognise the case change so I committed a change to rename completely and then another to make it correct ;)

isaacabraham (approved):

isaacabraham:
Thank you both very much!

PR #586, opened 2024-01-09
update fable version

Larocceau:
Update fable
* Ran into issue where did not compile F# 8 nested record update and replace syntax; newer fable version fixes this


Larocceau:
Still need to update minimal

PR #598, opened 2024-02-09
Upgrade to use metapackages

### LINKED ISSUE #557: Review declared dependencies of libraries that SAFE Template depends on
Triggered by discussion in #553.

We've had a few cases where upgrading packages causes problems because they don't declare the packages that they depend on tightly enough. In this case, Paket (or npm) upgrades packages to versions that it thinks should be compatible (and indeed that the packages themselves declare should work together), but the code fails at runtime because they are in fact incompatible.

We should review the packages that SAFE Template depends on, to check that they declare their dependencies correctly. If there are any problems, raise issues and/or PRs in the appropriate repo.

isaacabraham:
Closes #557. This PR introduces dependencies on two NuGet [metapackages](https://github.com/SAFE-Stack/SAFE.Meta), SAFE.Server and SAFE.Client. They contain the dependencies required for SAFE Stack applications as well as some nice helper types and functions.

* Only one dependency required for both Client and Server - less for end users to worry about.
* Opens the possibility of upgrades to existing SAFE Stack users "remotely" i.e. by releasing a new version of the metapackages - useful for e.g. security vulnerabilities etc.
* Allows people to more easily create their own SAFE Stack applications "from scratch".
* Adds some basic types and functions from e.g. Elmish Book as well as some helpers used internally at CIT.

I've also incorporated some of the types directly into the bookstore (notably, the AsyncOperation and Deferred types) (note that these will probably be renamed at some point to something a little more obvious).

I suggest that individuals test out this repository (or the SAFE.Meta packages) directly from this PR. Once we're happy with the shape of the metapackages, we can release it as a full v5 release (to go alongside SAFE Stack 5).


Content/default/src/Server/Server.fs
```diff
@@ -32,11 +30,7 @@ let todosApi = {
         }
 }
 
-let webApp =
-    Remoting.createApi ()
-    |> Remoting.withRouteBuilder Route.builder
-    |> Remoting.fromValue todosApi
-    |> Remoting.buildHttpHandler
+let webApp = Api.make todosApi
```
mattgallagher92, line 33: I would prefer Route.builder to be a required parameter, or for there to be an overload accepting it

Content/default/src/Client/Index.fs
```diff
@@ -1,43 +1,46 @@
...
+    | AddTodo msg ->
+        match msg with
+        | Start() ->
+            let todo = Todo.create model.Input
+            let cmd = Cmd.OfAsync.perform todosApi.addTodo todo (Finished >> AddTodo)
+            { model with Input = "" }, cmd
+        | Finished todo ->
+            {
+                model with
+                    Todos = model.Todos.Map(fun todos -> todos @ [ todo ])
```
mattgallagher92, line 41: It seems weird to me for `Map` to be a method rather than a function in the `Deferred` module.

Content/default/src/Server/Server.fs
isaacabraham: It's an optional parameter on `Api.make`. It defaults to `/api/{api name}/{api method}` which is a good starting point and most people tend to stick with.

Content/default/src/Client/Index.fs
isaacabraham: It's also a function in that module - I just left them as both functions and members. Happy to remove the members though!

Content/default/src/Server/Server.fs
isaacabraham: We should raise issues on the API design inside SAFE.Meta though.

Content/default/src/Client/Index.fs
isaacabraham: Ditto above.

isaacabraham:
@mattgallagher92 I'd like to get this merged in this week if possible - even if you have legitimate concerns about the API, we can rapidly iterate on that once the template is hooked up to using the package (and ideally published as a beta).

We can also "soft launch" the types and stuff (just including it in the package but not heavily publicising it) and iterating that way.

Content/default/src/Server/Server.fs
mattgallagher92: Great!

mattgallagher92 (approved):

PR #605, opened 2024-04-01
Upgrade to beta5

isaacabraham:


Content/default/tests/Client/Client.Tests.fs
```diff
@@ -12,19 +13,28 @@ let client =
...
+                 |> RemoteData.map List.head
+                 |> RemoteData.defaultValue (Todo.create ""))
+                newTodo
+                "Todo should equal new todo"
     ]
 
 let all =
     testList "All" [
-//-:cnd:noEmit
+        //-:cnd:noEmit
```
mattgallagher92, line 33: Don't indent these comments. It breaks the tests. See https://github.com/SAFE-Stack/SAFE-template/pull/581/files.

Perhaps it's worth a regular comment to say that?

Content/default/src/Client/Index.fs
```diff
@@ -5,89 +5,94 @@ open SAFE
...
     | SetInput value -> { model with Input = value }, Cmd.none
-    | GetTodos msg ->
+    | LoadTodos msg ->
         match msg with
         | Start() ->
-            let cmd = Cmd.OfAsync.perform todosApi.getTodos () (Finished >> GetTodos)
-            { model with Todos = InProgress }, cmd
-        | Finished todos -> { model with Todos = Resolved todos }, Cmd.none
-    | AddTodo msg ->
+            let loadTodosCmd = todosApi.getTodos () |> Async.toCmdUnsafe LoadTodos
```
mattgallagher92, line 31: I guess that, because `Async<'a>` is cold, the API isn't actually called until Elmish handles the command. Have we tested that?

I would feel weird about accidentally triggering an API call in the update loop as it's meant to be a pure function. Maybe it doesn't matter? Not sure what the implication is if there's a network exception or something.

Content/default/src/Client/Index.fs
```diff
@@ -5,89 +5,94 @@ open SAFE
...
+    | SaveTodo msg ->
         match msg with
-        | Start() ->
-            let todo = Todo.create model.Input
-            let cmd = Cmd.OfAsync.perform todosApi.addTodo todo (Finished >> AddTodo)
-            { model with Input = "" }, cmd
+        | Start todoText ->
+            let saveTodoCmd =
+                let todo = Todo.create todoText
+                todosApi.addTodo todo |> Async.toCmdUnsafe SaveTodo
```
mattgallagher92, line 39: I think that `Async.toApiCallCmd` might be a better name than `Async.toCmdUnsafe`. The current function implies to me more generic async behaviour than the specific `ApiCall` behaviour it actually represents.

In what sense is the function unsafe?

mattgallagher92 (changes requested):
The test change is a must and I think that the others are worth discussion.

Content/default/tests/Client/Client.Tests.fs
isaacabraham: This was done by Fantomas :-( Maybe we need to raise an issue for that?

Content/default/src/Client/Index.fs
isaacabraham: I spoke with @MangelMaxime about exactly this, and as you say, since they are cold workflow computations, this is safe to do. From a "pure" perspective it might be a little weird but from a "learner" perspective I think that this is easier to grok than the separate "method" + "arg" approach (plus pipelining seems more natural here).

Content/default/src/Client/Index.fs
isaacabraham: It doesn't handle exceptions. There's a `toCmd` variant which requires two messages e.g.

```fsharp
todosApi.addTodo todo |> Async.toCmd SaveTodo ApiError
```

or similar.

There's XML docs around this, but perhaps a different name?

I would like to encourage people to always include an error handler - too many errors get silently swallowed otherwise.

Content/default/src/Client/Index.fs
MangelMaxime: Personally, I think sticking to the native `Cmd` module is "better".

## Consistent API

This is because, like that people learn to work with `Cmd` in general and when they need to execute a function inside of `Cmd` they use the same API.

```fs
Cmd.OfAsync.attempt
Cmd.OfPromise.attempt
Cmd.OfFunc.attempt
// etc.
```

You always do:

1. I want a command : `Cmd`
2. I want to work with X (async for example): `Cmd.OfAsync`
3. I only case care about the success path: `Cmd.OfAsync.attempt`
4. Now I need to provide the "action", its arguments and the way to get back the result:

	```fs
	Cmd.OfAsync.attempt todosApi.getTodo () LoadTodos

	// ------------
	Cmd.OfAsync.attempt
		todosApi.getTodos
		()
		LoadTodos
	
	// ------------
	let args = { ... } // A really complex/lengthy arguments to build
	Cmd.OfAsync.attempt todosApi.getTodo args LoadTodos
	```

And these questions / logics can be applied to any type of commands.

## More focused intellisense

The intellisense also give them only the `Cmd` API not the whole `Async` API in the intellisense. Compare the result of `Cmd.OfAsync.*` vs `Async.*` where `*` is the place where you ask for intellisense.

Content/default/src/Client/Index.fs
```diff
@@ -5,89 +5,94 @@ open SAFE
...
+    | LoadTodos of ApiCall<unit, Todo list>
+    | SaveTodo of ApiCall<string, Todo>
 
 let todosApi = Api.makeProxy<ITodosApi> ()
 
 let init () =
-    let model = { Todos = NotStarted; Input = "" }
-    model, Cmd.ofMsg (GetTodos(Start()))
+    let initialModel = { Todos = NotStarted; Input = "" }
+    let initialCmd = Cmd.ofMsg (LoadTodos(Start()))
```
MangelMaxime, line 21: In these situation in general, I personally prefer to use pipeline:

```fs
Start
|> LoadTodos
|> Cmd.ofMsg

// Real world example

Router.OrderRoute.Index
|> Router.AuthRoute.Orders
|> Router.Route.Auth
|> Router.newUrl
```

Content/default/src/Client/Index.fs
isaacabraham: Regarding the ApiCall behaviour - it will actually work with any Async workflow, although you're right, it only works with the `ApiCall` type as it automatically wraps it in `Finished`. I also considered:

`ApiCall.dispatch`
`ApiCall.toCmd` (but this is misleading to me, we're not going from an ApiCall to a Cmd)

Your suggestion of `Async.toApiCallCmd` is the most accurate, but it feels like a bit of a mouthful?

Content/default/src/Client/Index.fs
isaacabraham: These are good points. Perhaps we can piggy back on the Cmd module instead e.g.

`Cmd.ofAsync.startApiCall` or similar?

The concerns I have with the `Cmd.OfAsync` module are:

1. `attempt` and the other names are sub-optimal - they do not easily reflect the behaviours that distinguish each of them.
2. The separate "method" and "argument" approach is confusing. People's initial inclination is to simply want "here's an Async, please can you map it to a command".

I've seen beginners all fail to grasp the `Cmd.OfAsync` module in this way - and the same way that the different modules e.g. `OfAsync` rather than `OfAsyncWith` (or whatever it's called) is also not immediately obvious.

I guess I'm trying to provide a simple, constrained model that fits for the beginner and (ideally) 80-90% of the most common use cases people have, especially with Fable Remoting.

Content/default/src/Client/Index.fs
isaacabraham: How about `Cmd.OfAsyncApiCall.xxx` where `xxx` is some meaningful name e.g.

* `execute`
* `start`
* `executeSafe`

Open to ideas :)

Content/default/src/Client/Index.fs
MangelMaxime: Personally, I don't find `startApiCall` to be a better name because `startApiCall` feels limited to API call when it works for any async operations. 

`Async.toCmdUnsafe` was created to kind of escape the `Cmd` knowledge but it back into it because people need to find the `*Cmd*` API in the whole intellisense proposition.

I used Fable.Remoting in one of my project with a beginner client and using `Cmd.OfAsync` never caused issues, once you know the pattern you always apply the same code.

IHMO in this situation it is best to teach people how Elmish API works and make them understand it than adding a layer of abstraction on top of it. My reason for that is I often see people who used abstraction and are lost once they need to think a little bit out of the box / standard canvas that was teaches to them.

The way I teach it to people is by either looking at the signature of the function:

1. Do you need to be notified on success and don't care about error? `Cmd.attempt`
2. Do you need to be notified only for errors? `Cmd.perform`
3. Do you want to be notified in both cases? `Cmd.either`
4. Do you just want to "execute" it and never be notified then use `Cmd.exec` => This is one doesn't exist in Elimish standard API, I always add it in the project manually.

Coming back to them being lost, this is also why I teach/mention early that `init`, `update` and `view` are just normal functions. They requires in general 2 arguments but if needed please use 3,4,... arguments for theses functions.

I don't think the complexity of mastering Elmish is coming from the `Cmd` API, but more in finding a way to teach people to create re-usable components or patterns in how to structure their code / modules in a way that the code is easily discoverable.

MangelMaxime: Another ways I structure API in similar situation is to create a completely independent module which allows me to hide implementation details, not have existing naming convey a meaning and have a minimal intellisense.

For example,

```fs
// Generate `Cmd` under the hood to navigate to a new URL from the update function
Router.OrderRoute.Index
|> Router.AuthRoute.Orders
|> Router.Route.Auth
|> Router.newUrl

// Generate `Cmd` under the hood to trigger a toast from any update function
Toast.message "Hello, I am a toast"
|> Toast.position Toast.BottomLeft
|> Toast.timeout (TimeSpan.FromSeconds (3.0))
|> Toast.info

// API calls in my different projects often looks like
// Public APIs
Http.get "/some/url"
Http.get "/some/url" bodyData

// Authenticated APIs
Http.Auth.get session "/some/url"
Http.Auth.post session "/some/url" bodyData

// etc.
```

mattgallagher92: Maxime, you can see in the [implementation code](https://github.com/SAFE-Stack/SAFE.Meta/blob/main/src/SAFE.Client/SAFE.fs#L40-L98) that this is wrapping a successful response in the `Finished` case of the `ApiCall` type.

>People's initial inclination is to simply want "here's an Async, please can you map it to a command".

Having read that line @isaacabraham, I get the motivation for this change. You're expecting people to behave like this: "I want to call `todosApi.getTodos`, I guess I do `todosApi.getTodos ()`, but now I have an async :thinking: maybe there's something in the `Async` module that can help me...". However, I'm not sure how likely it is that they'll immediately reach for the `Async` module, and the template already provides an example of how to work with an API, which I imagine will unblock people pretty quickly.

On the whole I agree with Maxime's points. In particular, I agree that adding a layer of abstraction makes it harder to break out of that when necessary. It's also a very shallow abstraction - `Async.toCmdUnsafe (f x) g` is the same as `Cmd.OfAsync.attempt f x (Finished >> g)` - which I generally find to be a bit of a code smell.

I actually like the function and arguments being separate passed separately - it feels like I'm telling Elmish "here are the ingredients you need to make an async when you're ready" and I know that the async won't be started immediately. If the async is awaiting a task then it could otherwise be triggered immediately. IMO, that's a subtlety that will be much harder for newcomers to grok than passing functions and arguments separately.

That said, I also agree Isaac that the names could be improved. Maybe that's best fixed by adding new functions to Elmish that pass through to the old ones and/or contributing a PR that improves the doc comments?

Content/default/src/Client/Index.fs
mattgallagher92: I'm currently thinking that we just shouldn't include this function at all :sweat_smile: (see other thread)

Happy to consider names if/when we agree to add this.

Content/default/src/Client/Index.fs
mattgallagher92: Maybe we should add a comment in the template linking to the relevant recipe(s), e.g. https://safe-stack.github.io/docs/recipes/client-server/messaging/ and update the recipe(s) if necessary?

mattgallagher92: To highlight that the names could be improved, these are the wrong way round.

>1. Do you need to be notified on success and don't care about error? `Cmd.attempt`
>2. Do you need to be notified only for errors? `Cmd.perform`

isaacabraham: I like the idea of adding links to recipes to the xmldocs - this could really help beginners. Maybe we actually need a recipe / explanation post on the different `Cmd` modules e.g. `Async` etc..

And yes, the abstraction is very shallow, it doesn't do much except wrap the `Finished` thing. I have an alternate version that was using a class rather than functions - the benefit being that there was only one member (either overloaded or with an optional argument for the `onError` message.) which meant that there was nothing to learn in terms of member names e.g.

```fsharp

Cmd.ofAsync.Execute(api.getTodos (), GetTodos) // don't handle server exceptions
Cmd.ofAsync.Execute(api.getTodos (), GetTodos, HandleServerException) // also valid
```

isaacabraham: > I actually like the function and arguments being separate passed separately - it feels like I'm telling Elmish "here are the ingredients you need to make an async when you're ready" and I know that the async won't be started immediately. If the async is awaiting a task then it could otherwise be triggered immediately. IMO, that's a subtlety that will be much harder for newcomers to grok than passing functions and arguments separately.

I know that is a subjective thing - my opinion isn't the "one true way" :) I'm not convinced though that the current API can't be improved but perhaps I'm in the minority here :) My main concern is that outside of Elmish, this pattern of "pass a function and then pass it's arguments separately" is very, very unusual - the only ways I've really seen this done in the past is with a thunk e.g. `(fun () -> api.GetTodos())` but in the case of Async this isn't even needed since it's a cold computation. I'm not sure about the Task stuff either - there are no Tasks in JS land, and even if it did, wouldn't that only apply if you started the task *outside* of the Async block?

I'm really just focusing of the use case for 80-90% of the time: "I have an API call that I want to make and I want the result to be pushed into this Elmish message when it returns" and optimising that as much as possible. Perhaps that's not the right way to go about this though...

mattgallagher92: Yeah, you'd need the task to be started (hence created) separately for the hot start to be a problem, for example `someTask |> Async.AwaitTask`. As you say, given we're working in the browser, maybe that's not actually something that we need to worry about - it's presumably all promises that cold start? My JS knowledge is a bit hazy here tbh. Let's set that aside for the moment and assume that it's fine.

I get what you're saying with it being unusual. And I was thinking that people might need to use the old API a lot, which would significantly weaken the case for the new API (and provide an argument against it), but on reflection maybe that's not true. I think that there probably is a good case for something like the change that you're proposing.

I think most of my remaining reservation comes from the fact that the `Finished` part of the `ApiCall` is hidden in the function called from the `Start` handler. There's a weird asymmetry there that makes it feel like we're not working completely at either the `Async` level of abstraction or the `ApiCall` level. How about we just use asyncs and don't do any `ApiCall` stuff at all in the background?

```suggestion
            let loadTodosCmd = todosApi.getTodos () |> Async.toCmdUnsafe (Finished >> LoadTodos)
```

I'm okay with that and it doesn't separate the argument and function, which seems to be the main motivator. Then I think the last thing we'd need to do would be to settle on an appropriate name and module.

Content/default/tests/Client/Client.Tests.fs
mattgallagher92: cc @nojaf

Content/default/src/Client/Index.fs
mattgallagher92: Thanks Maxime :slightly_smiling_face: The point about separate messages is an interesting one - I like the clarity you get that the two messages are linked, and it potentially halves the number of cases in `Msg`, which can be significant, but the flat `Msg` structure is a little bit simpler. I don't see a strong argument either way, so happy to leave it as is for now.

Isaac, as another data point, I discussed your proposed change with @Larocceau who said that he didn't have any trouble with Cmd.OfAsync.perform when first encountering it as an F# novice - checking the type signature was enough to quickly work out what was going on. On that basis, I don't really see how the slightly unconventional nature is a problem - new F# devs don't have expectations and experienced F# devs can check the signature and easily understand what's necessary.

mattgallagher92 (approved):

PR #613, opened 2024-06-07
Do `dotnet tool restore` automatically in build target

theprash:
By adding a target in Build.fsproj to run `dotnet tool restore`, we can remove the step that requires users to do it manually.

Should we go ahead with this? Discussion welcome.

### DIFF
--- a/Build.fsproj
+++ b/Build.fsproj
@@ -12,4 +12,7 @@
   </ItemGroup>
 
   <Import Project=".paket\Paket.Restore.targets" />
+  <Target Name="DotNetToolRestore" BeforeTargets="PaketRestore">
+    <Exec Command="dotnet tool restore" />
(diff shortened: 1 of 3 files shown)


mattgallagher92 (commented):
Looks great to me!

It'll add a very small amount of time on every build, but I think that the benefits (better experience for new devs; automatic tool restore when switching branches) are worth it 😀

mattgallagher92:
We likely want to update the docs site too.

mattgallagher92:
Maybe we want to change the root level Build.fsproj to be consistent?

Larocceau:
@mattgallagher92 also added the automatic restore to the root level build.fsproj. I think we can merge this?

mattgallagher92 (approved):

Issue #615, opened 2024-06-15
Failing Test


giuliohome:
Fable, you have a template with a failing test—why?!

```
dotnet new install SAFE.Template
dotnet new SAFE
dotnet tool restore
dotnet run -- RunTests
```

![immagine](https://github.com/SAFE-Stack/SAFE-template/assets/3272563/eb67e525-49a7-4b0f-9bb5-fa54e43c9de4)

https://github.com/SAFE-Stack/SAFE-template/blob/e8677f329079b5741f587289526b9fe18f00675b/Content/default/tests/Client/Client.Tests.fs#L16)

Is one supposed to fix it this way? 

```fsharp
let model, _ = update (LoadTodos (Finished [newTodo])) model
```

or if you prefer

```fsharp
let model, _ = update (LoadTodos (Finished [])) model
let model, _ = update (SaveTodo(Finished newTodo)) model
```

Is it a template or an examination/test?

### Comment about using Copilot

Copilot and GPT cannot pass it.

![immagine](https://github.com/SAFE-Stack/SAFE-template/assets/3272563/f47c04d2-00e3-4141-8f63-7cd3840ddcf0)

I'm happy that Copilot cannot do that because it means human programmers' jobs are saved! 😄 

![immagine](https://github.com/SAFE-Stack/SAFE-template/assets/3272563/55022d7f-a2d6-494e-80f1-d6ef7c313bf2)

But Copilot actually helps because after the above changes you will see that the root cause error is "Todos were not in the Loaded state".

![immagine](https://github.com/SAFE-Stack/SAFE-template/assets/3272563/a54dbd85-40ed-49dd-9669-4fb02b5711a1)

But Copilot actually helps because, after making the above changes, you will see that the root cause of the error is 'Todos were not in the Loaded state,' even though it is not really capable of producing the correct fix for the problem.

giuliohome:
Separate question.

Besides fixing the test (as above), one must first understand that server project tests appear in the console output, whereas client project tests appear in the browser. This can be quite confusing for beginners and turns it into a specialized template. For experts too, it might be convenient, if not necessary, to get the return code of the client tests (and possibly the output) from the console as well as from the browser. Would it be possible to modify the template to achieve this?"

isaacabraham:
@giuliohome If I remember correctly, the test deliberately fails to show you how a failing test will appear in the app! But I'll double check.

Larocceau:
[...]

PR #616, opened 2024-06-16
Fix bug issue #615 in Client.Tests.fs

giuliohome:
Fix bug issue #615

Helping Copilot to explain the fix, **ex-post**.

> giuliohome: explain why adding line "let model, _ = update (LoadTodos (Finished [])) model" fix a bug so that the test does not fail now

>GitHub Copilot: The line `let model, _ = update (LoadTodos (Finished [])) model` is loading an empty list of todos into the model. This is likely fixing a bug where the model's todo list was not initialized properly, causing the test to fail.

> In the previous state, the model might have been in an uninitialized state or contained some stale data. By explicitly loading an empty list of todos, it ensures that the model is in a known and clean state before the test starts to add a new todo. This makes the test more reliable and predictable, as it's not affected by the previous state of the model. 

> This is a common practice in unit testing, where each test should be isolated and not depend on the state from other tests. It's also part of the Arrange-Act-Assert pattern, where you first arrange (set up) the system under test, then act on it, and finally assert that it behaved as expected. In this case, loading the empty list of todos is part of the arrange phase.

**Now in my words.** 

The issue arises because the `init` function generates a model with todos that have not started yet. As a result, the remote `map` used for the `Length` assertion doesn't map to any function, causing the test to fail.

### DIFF
--- a/Content/default/tests/Client/Client.Tests.fs
+++ b/Content/default/tests/Client/Client.Tests.fs
@@ -12,7 +12,7 @@ let client =
         <| fun _ ->
             let newTodo = Todo.create "new todo"
             let model, _ = init ()
-
+            let model, _ = update (LoadTodos (Finished [])) model
             let model, _ = update (SaveTodo(Finished newTodo)) model
 
             Expect.equal


giuliohome:
Note that this PR does not address the separate question asked here: https://github.com/SAFE-Stack/SAFE-template/issues/615#issuecomment-2169434320

Larocceau:
@giuliohome thanks for your contribution, it has been merged and [release](https://github.com/SAFE-Stack/SAFE-template/releases/tag/5.0.6). We'd like to keep the discussion #615 going to determine whether we'd like this to be the final solution, or make the demo app slightly more resilient by making sure the "add todo" operation can work even if loading the todos somehow failed.

Issue #618, opened 2024-06-18
Use `npm ci` rather than `npm install` in Build.fs?


mattgallagher92:
`npm ci` would restore dependencies specified in package-lock.json (if consistent with package.json), but not alter it. See https://docs.npmjs.com/cli/v10/commands/npm-ci. It might be faster too.

If developers want to alter what's in the lockfile, it feels like they should be deliberately running an `npm install`, rather than it being a side effect of build targets.

Any reason not to do this?

mattgallagher92:
It seems that it's much slower! Let's hold off for now.

jwthomson:
I am quite confident that `npm ci` is the preferred command for build pipelines.

> It seems that it's much slower!
Where have you observed this?

I think it is better to be slow (hopefully this can be rectified) and safe than fast and potentially "dangerous".

For now I have approved this in #619 and it has been merged but we can always undo this if the slowdown is very significant and canot be resolved.

mattgallagher92:
Prash tested it and said that it was of the order of 10 seconds (may even have been 20) to run `npm ci` as compared to less than 1 second for `npm install`

Larocceau:
Sorry for merging this! I did not realize @jwthomson was not aware of the discussion we had on this last week

jwthomson:
I've run each a few times locally and it is slower, at least on my machine. It was a lot slower the first time. It would be useful to test this in a real dev scenario where npm ci is being run fairly often but not lots of times in immediate succession, to get a more accurate feel for how this would work in practice.

|| `npm ci` | `npm install` |
|-|-:|-:|
| Cold, first run | 14.9s | 1.2s |
| Successive runs | ~4s | ~1.2s |

jwthomson:
Following some discussion around caching, I have re-run these tests making sure that they are both using fresh, empty cache directories.

I'm measuring using the following:

```powershell
Measure-Command { npm install --cache C:\tmp\install-cache }
Measure-Command { npm ci --cache C:\tmp\ci-cache }
```

I think it is important to look at the relative times rather than the raw times. Overall I don't there's a significant difference vs last week's tests.

|| `npm ci` | `npm install` |
|-|-:|-:|
| Cold, first run (empty cache) | 7.1s | 0.8s |
| Successive runs, without clearing cache (mean of 3) | 2.6s | 0.8s |

isaacabraham:
I think we've agreed to go ahead with this - can be closed once the latest template is released.

[...]

PR #619, opened 2024-06-21
Use npm ci in the build script so package.lock cannot get updated there

Larocceau:
addresses #618

* NPM ci only installs packages according to package.json and package.lock, failing if they don't line up. CI is faster and will prevent unexpected changes to package.lock until the developer explicitly uses NPM install

### DIFF
--- a/Content/default/Build.fs
+++ b/Content/default/Build.fs
@@ -20,7 +20,7 @@ Target.create "Clean" (fun _ ->
     run dotnet [ "fable"; "clean"; "--yes" ] clientPath // Delete *.fs.js files created by Fable
 )
 
-Target.create "InstallClient" (fun _ -> run npm [ "install" ] ".")
+Target.create "InstallClient" (fun _ -> run npm [ "ci" ] ".")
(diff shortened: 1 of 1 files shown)


Content/default/Build.fs
```diff
@@ -20,7 +20,7 @@ Target.create "Clean" (fun _ ->
     run dotnet [ "fable"; "clean"; "--yes" ] clientPath // Delete *.fs.js files created by Fable
 )
 
-Target.create "InstallClient" (fun _ -> run npm [ "install" ] ".")
+Target.create "InstallClient" (fun _ -> run npm [ "ci" ] ".")
```
Larocceau, line 24: Rename to RestoreCLientPackages

Larocceau:
install seems to be quicker than ci 🤯. Let's check again?

Content/default/Build.fs
jwthomson: I've made this change

jwthomson (approved):

PR #623, opened 2024-07-12
Run project tests as part of template testing

Larocceau:
* Add build action in default project to run tests one-off using mocha test runner (instead of in browser)
* Add build action in template build script to run these tests

### DIFF
--- a/Build.fs
+++ b/Build.fs
@@ -87,7 +87,7 @@ Target.create "TemplateExecutionTests" (fun _ ->
 
 Target.create "DefaultTemplateTests" (fun _ ->
     let cmd = "run"
-    let args = "tests --project Build.fsproj"
+    let args = "RunTestsHeadless --project Build.fsproj"
     let result = DotNet.exec (fun x -> { x with DotNetCliPath = "dotnet"; WorkingDirectory = defaultTemplatePath}) cmd args
     if not result.OK then failwithf "`dotnet %s %s` failed" cmd args
 )
--- a/Build.fs
+++ b/Build.fs
@@ -87,7 +87,7 @@ Target.create "TemplateTests" (fun _ ->
 
 Target.create "DefaultProjectTests" (fun _ ->
     let cmd = "run"
-    let args = "tests --project build.fsproj"
+    let args = "tests --project Build.fsproj"
     let result = DotNet.exec (fun x -> { x with DotNetCliPath = "dotnet"; WorkingDirectory = defaultProjectPath}) cmd args
     if not result.OK then failwithf "`dotnet %s %s` failed" cmd args
 )
--- a/Build.fs
+++ b/Build.fs
@@ -12,7 +12,7 @@ let skipTests = Environment.hasEnvironVar "yolo"
 let release = ReleaseNotes.load "RELEASE_NOTES.md"
 
 let templatePath = "./Content/.template.config/template.json"
-let defaultTemplatePath = "./Content/default"
+let defaultProjectPath = "./Content/default"
 let templateProj = "SAFE.Template.proj"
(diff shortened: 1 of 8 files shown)


azure-pipelines.yml
mattgallagher92, line 1: What's the reason for changing to double quotes?

Content/default/Helpers.fs
```diff
@@ -77,16 +77,22 @@ let createProcess exe args dir =
...
         | Some path -> path
         | None ->
             "npm was not found in path. Please install it and make sure it's available from your path. "
             + "See https://safe-stack.github.io/docs/quickstart/#install-pre-requisites for more info"
             |> failwith
 
-    createProcess npmPath args dir
+    createProcess path args dir
+
+let npm: seq<string> -> string -> CreateProcess<ProcessResult<unit>> =
```
mattgallagher92, line 91:
```suggestion
let npm : seq<string> -> string -> CreateProcess<ProcessResult<unit>> = 
```

Formatting nitpick. Perhaps worth running fantomas?

Content/default/Helpers.fs
mattgallagher92, line 1: Why bother with the type annotations?

Content/default/Build.fs
```diff
@@ -71,6 +81,7 @@ let dependencies = [
 
     "Clean" ==> "RestoreClientDependencies" ==> "Run"
 
+    "restoreClientDependencies" ==> "Tests"
```
mattgallagher92, line 84:
```suggestion
    "RestoreClientDependencies" ==> "Tests"
```

Build.fs
mattgallagher92, line 1: I'm not sure that the terminology is quite precise enough here. I'd go for "default _template_ tests" and "template _execution_ tests".

Content/default/Build.fs
mattgallagher92, line 1: Maybe to make the distinction clearer, we should call these targets "RunTestsHeadless" and "WatchRunTests"?

If we rename "RunTests" we should also update docs, so I'm happy for that to be split into a separate issue/PR rather than done now.

mattgallagher92 (changes requested):
This looks like a good step in the right direction 🙂 I've made a few suggestions

Content/default/Helpers.fs
Larocceau: Yeah I get this weird error about these functions being generic? I don't understand how, but the annotations fix it

azure-pipelines.yml
Larocceau: OH YIKES, my ide did that. Will revert

Content/default/Build.fs
Larocceau: did the rename for runTestsHeadless, let's do the other change next week in conjunction with a docs update

Larocceau:
All feedback dealt with

Content/default/Build.fs
mattgallagher92: Thanks! Could you raise an issue to make sure that we remember to do that?

Content/default/Helpers.fs
mattgallagher92: Let's discuss.

Content/default/Build.fs
mattgallagher92: Let's make sure to update the docs too 👍🏼

mattgallagher92 (approved):
Happy for this to be merged so long as the docs are updated to reflect the changed target name

mattgallagher92 (changes requested):
Oh, one last thing: the template's README needs updating to reflect the changed target name.

mattgallagher92 (approved):

mattgallagher92:
My last comment is wrong - the README had already been updated.

Issue #617, opened 2024-06-17
Add exclude files to settings.json


isaacabraham:
We should consider excluding folders from VS Code. This can be included in the `settings.json` file:

```json
    "files.exclude": {
        ".vs": true,
        ".npmrc": true,
        ".paket": true,
        "**/bin": true,
        "**/obj": true,
        "**/output": true,
        "deploy": true,
        "node_modules": true,
        "paket-files": true,
        ".fantomasignore": true,
        "src/Client/postcss.config.js": true,
        "src/Client/vite.config.mts": true,
        "src/Client/tailwind.config.js": true,
        "src/Client/index.css": true,
        "src/Client/index.html": true,
    }
```

This is just an idea - the list above might be too much though. Certainly the first half of the list would be good to add - the second half might be more controversial ;-)

Before:

![image](https://github.com/SAFE-Stack/SAFE-template/assets/1781813/6de22b25-4ac5-4e40-a1fe-79ea36b28c11)

After:

![image](https://github.com/SAFE-Stack/SAFE-template/assets/1781813/698920d7-5b9c-4e1a-8cbf-2fdb2b437333)

Larocceau:
I agree on excluding the first part of the list, second part not so much. I feel like those files tend to be a source of issues that can be fairly hard to pinpoint because they are in config files. For people not as familiar with the template hiding part of the config can make troubleshooting more challenging that necessary.

jwthomson:
@isaacabraham what are your thoughts on Maxime's preferred solution to this problem - nesting config under a single file using this extension:
https://github.com/antfu/vscode-file-nesting-config

isaacabraham:
I'm open to that. The other issue with my solution is that it doesn't just hide it from the explorer, it hides it from CTRL + P file navigation as well.

isaacabraham:
I'm just not sure where we would put / group all these files / folders under.

mattgallagher92:
I don't use VS Code, so won't make the change, but my opinion is that we should hide build assets/output and nest infrequently-used files that affect the behaviour of the build and/or tools. Hiding config files would make things more magical and harder to debug.

We should exclude these files:
  - .vs (not relevant to VS Code users)
  - .paket
  - **/bin
  - **/obj
  - **/output
  - deploy
  - node_modules
  - paket-files

And we should nest these files:
  - .npmrc (under package.json)
  - .fantomasignore (under .gitignore???)
[...]

PR #625, opened 2024-08-16
Hide project / output / cache files from VS Code

jwthomson:
Relates to #617.

After some discussion we thought that ignoring some files is a good idea, but hiding files as a tree under other files could be confusing. Most of the files which were proposed for hiding in a tree were config files which are more likely to need to be edited so we opted not to nest them.


isaacabraham:
@jwthomson is this good to go?

Larocceau (changes requested):
check if this can be done with the gitignore instead

Larocceau:
@isaacabraham, no; we later decided to see if we can use the gitignore for this. I redacted my approval

jwthomson:
There were some files which are ignored in the `.gitignore` which someone may wish to edit from within VS code, e.g. user-specific settings files. So I opted **not** to persue using the `.gitignore` file as a basis for what files should be hidden.

Larocceau (approved):

PR #632, opened 2024-10-11
upgrade to Safe.Meta 5.1.0

### LINKED ISSUE #631: Upgrade to Meta v 5.1.0
We released meta 5.1.0 last week, so the dependency can be updated.

We should have a good look at implementing the updated RemoteData type correctly:
* Use the StartLoading function when refreshing, so existing data is not erased
* Use ToOption to extract the data when displaying, so there is no flicker (this was the main aim of the change)

Larocceau:
Upgrade to use [SAFE.Meta 5.1.0](https://github.com/SAFE-Stack/SAFE.Meta/releases/tag/5.1.0)
Closes #631


Content/default/paket.dependencies
mattgallagher92, line 1: You don't need to change these lines. `~>` means "compatible with", so `~> 5` is fine.

See https://fsprojects.github.io/Paket/nuget-dependencies.html#Pessimistic-version-constraint.

Larocceau: Ah crap, I was afraid of that... that means ppl are going to have unexpected breaking changes when they update their packages on SAFE 5.x apps. Perhaps we should release a 5.1.1 of the meta packages and revert those breaking changes? Then we could release them as 6.0.0, paired with SAFE.Template 6.0.0?

mattgallagher92: Hmm yeah good point 🫤 What do you reckon about a 6.0 release @isaacabraham?

mattgallagher92: If we don't want a v6 template release, we could break the synchronisation with metapackages.

jwthomson (approved):

PR #639, opened 2024-11-29
Update dependencies

Larocceau:
* run dotnet paket update --keep-major
* run npm update on both templates


Larocceau:
this replaces #635, and should solve the problem brought up by @OutOfScopeia where his installation on a fresh VM fails

jwthomson (approved):

Issue #626, opened 2024-08-23
Change Build.fs to explicitly restore


isaacabraham:
In light of #621 I think that we should change the run process:

1. Do a standard `dotnet restore` 
2. Do a `dotnet watch run --no-restore`

This has two benefits:

1. It solves this weird issue that some people are seeing with watch run.
2. It will give a small perf improvement to watch reloads because the restore process is skipped on restarts.

jwthomson:
I am looking into this

isaacabraham:
I'm reopening this because of my comments in #634 and assigning it to @Larocceau .

PR #640, opened 2024-11-29
Fix error caused by restore on running tests + build solution before run, instead of just restore

Larocceau:
* Remove dotnet restore action before running the watcher
* instead of building just the shared path, build the full solution
* Use `no-restore` when running server tests
* restore full solution before running any tests

addresses what is left of #626 after partial fix in #634

### DIFF
--- a/Content/default/Build.fs
+++ b/Content/default/Build.fs
@@ -74,16 +74,11 @@ Target.create "Format" (fun _ -> run dotnet [ "fantomas"; "." ] ".")
 open Fake.Core.TargetOperators
 
 let dependencies = [
-    "Build" ==> "RunTestsHeadless"
-    "Build" ==> "WatchRunTests"
-    "Build" ==> "Run"
-
     "Clean" ==> "RestoreClientDependencies" ==> "Bundle" ==> "Azure"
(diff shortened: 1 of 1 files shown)


Content/default/Build.fs
```diff
@@ -78,6 +74,10 @@ Target.create "Format" (fun _ -> run dotnet [ "fantomas"; "." ] ".")
 open Fake.Core.TargetOperators
 
 let dependencies = [
+    "Build" ==> "RunTestsHeadless"
+    "Build" ==> "WatchRunTests"
+    "Build" ==> "Run"
```
isaacabraham, line 79: Could you merge this with the Run pipeline below?

isaacabraham:
Looks good to me overall :-)

Content/default/Build.fs
Larocceau:
```
let dependencies = [
    "Clean" ==> "RestoreClientDependencies" ==> "Bundle" ==> "Azure"
    "Clean" ==> "RestoreClientDependencies" ==> "Build" ==> "Run"

    "RestoreClientDependencies" ==> "Build" ==> "RunTestsHeadless"
    "RestoreClientDependencies" ==> "Build" ==> "WatchRunTests"
]
```
Like so?

PR #642, opened 2024-12-24
Bumping node version to restriction latest 22

halcwb:
When using the latest stable node version, the template cannot be used. Updated the restriction on the node version to 22.


jwthomson:
Thanks for submitting this PR, it is much appreciated 😄

Can I suggest rather than using the `||` oerator, can you use a range like this: `"node": ">= 18 <= 22"` instead? Otherwise this will be a breaking change for people on Node version 20. This is probably the syntax we should have used in the first place, so apologies for providing a bad template for the changes!

PR #644, opened 2025-01-24
Automated relase note file update

Larocceau:
* automatically write release notes after release

### DIFF
--- a/Build.fs
+++ b/Build.fs
@@ -98,6 +98,23 @@ Target.create "Release" (fun _ ->
     let result = DotNet.exec (fun x -> { x with DotNetCliPath = "dotnet"; WorkingDirectory = nupkgDir }) "nuget" nugetArgs
     if not result.OK then failwithf "`dotnet %s` failed" "nuget push")
 
+Target.create "UpdateReleaseNotes" (fun _ ->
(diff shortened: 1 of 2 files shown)


Build.fs
```diff
@@ -98,6 +98,23 @@ Target.create "Release" (fun _ ->
     let result = DotNet.exec (fun x -> { x with DotNetCliPath = "dotnet"; WorkingDirectory = nupkgDir }) "nuget" nugetArgs
     if not result.OK then failwithf "`dotnet %s` failed" "nuget push")
 
+Target.create "UpdateReleaseNotes" (fun _ ->
+    let file = "RELEASE_NOTES.md"
+    let version = Environment.environVarOrFail "VERSION"
+    let body = Environment.environVarOrFail "RELEASE_NOTES_BODY"
+
+    let existingContent = System.IO.File.ReadAllText file
```
jwthomson, line 106: It is unfortunate that our changelog has the newest changes at the top because prepending to a file is more of a pain than appending.

Obviously with this small text file it is fine but in 1000 years when the changelog is trillions of GB it won't be a good idea to slurp up the whole file into memory (I suppose by then we might have enough RAM)

PR #649, opened 2025-03-28
Node 22 support

jjbrooksnet:
Modified version of Casper Bollen's changes to support the current LTS version of Node (v22) without losing support for Node v20


Content/default/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "private": true,
     "engines": {
-        "node": "~18 || ~20",
+        "node": ">= 18 <= 22",
         "npm": "~9 || ~10"
```
Larocceau, line 5: Node 18 will lose support later this year. I would suggest changing this to 20 || 22 to make sure we only support current LTS versions

jjbrooksnet: We should drop Node 18 support only in the next major release. But I have changed this to support 18, 20 and 22 instead of the non-LTS releases between 18 and 22

Larocceau (approved):

